PR #29, opened 2017-12-12
wait for file_server ready to avoid timing issue

### LINKED ISSUE #28: Testcase TestObtainFromKnownLocation intermittently failed as file_server was not ready
It seems like a timing issue. The code started the file server in a background goroutine. The server might not be ready when the code tried to connect.

func testArchiveSource(t *testing.T, tc obtainTestCase) {
	cleanup(tc.targetDir)
	defer cleanup(tc.targetDir)
	**server := testCommon.StartStaticFileServer(t)**
	defer testCommon.ReportOnError(t, func() error { return server.Shutdown(context.TODO()) })
	source := NewArchiveSource(tc.url, tc.targetDir)
	runner := test.NewMockRunner()
	runner.PrepareCommandExpectation(tc.expectedCommands)
	fs := runner.GetFileSystem().(*test.MockFileSystem)
	fs.PrepareChdir(tc.expectedChdir)
	if tc.getWdErr != nil {
		fs.On("Getwd").Return("home", *tc.getWdErr).Once()
	}
	**err := source.Obtain(runner)**

northtyphoon:
**Purpose of the PR:**

**Fixes #28** 

**Notes/Details:**

The fix pings the file server to test the connection after it's started.

@sajayantony @shhsu

### DIFF
--- a/tests/testCommon/file_server.go
+++ b/tests/testCommon/file_server.go
@@ -3,11 +3,13 @@ package testCommon
 import (
 	"archive/tar"
 	"compress/gzip"
+	"fmt"
 	"io"
 	"net/http"
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 )
 
 const staticFilePort = ":32764"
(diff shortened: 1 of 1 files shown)


tests/testCommon/file_server.go
```diff
@@ -29,9 +31,30 @@ func StartStaticFileServer(t *testing.T) *http.Server {
...
+	err := WaitForServerReady(10)
+	if err != nil {
+		t.Fail()
+	}
+
 	return server
 }
 
+// WaitForServerReady waits for server ready until timeout
+func WaitForServerReady(timeoutInSeconds uint) error {
```
shhsu: After some quick research, apparently the right thing to do is not use ListenAndServe at all.
http://grokbase.com/t/gg/golang-nuts/141ee4dqyg/go-nuts-how-to-know-when-listenandserve-is-ready-to-handle-connections
Would you like to modify with this solution?

northtyphoon: Thought about it, current solution based on client pulling also works for ListenAndServeTLS if we want to test https in the future.

Issue #25, opened 2017-12-07
Output digests of build dependencies


shhsu:
Current output format:
```json
[{"image": "<image>", "build-dependencies": ["<dep1>:<tag1>", "<dep2>:<tag2>"], "runtime-dependency": "<base1>:<tag3>"}, ...]
```

Proposed change

```
image : 
    repo , [tag], digest
    -> deps [
        { repo [tag] digest , type },
        {  repo [tag] digest , type }
```

PR #31, opened 2018-01-19
Output digests of build dependencies

shhsu:
**Purpose of the PR:**
Output digests of build dependencies.

**Fixes #** 
#25


northtyphoon:
@ehotinger

pkg/commands/docker.go
```diff
@@ -125,3 +123,33 @@ func (t *dockerBuildTask) Push(runner build.Runner) error {
...
+			}
+		}
+	}
+	return nil
+}
+
+func queryDigest(runner build.Runner, reference *build.ImageReference) error {
+	refString := reference.String()
+	line, err := runner.QueryCmd("docker", []string{
+		"image", "ls", "--digests", "--format", "\"{{ .Digest }}\"", refString,
```
northtyphoon: {{ .Digest }} what is the extra space for?

pkg/grok/docker_compose.go
```diff
@@ -113,15 +113,19 @@ func ResolveDockerComposeDependencies(env *build.BuilderContext, projectDirector
 		} else {
 			dockerfilePath = filepath.Join(imageContext, env.Expand(service.Build.Dockerfile))
 		}
-		runtime, buildtime, err := ResolveDockerfileDependencies(dockerfilePath)
+		runtime, buildtimes, err := ResolveDockerfileDependencies(dockerfilePath)
```
northtyphoon: Nit: ImageDependencies use Buildtime, better be consistent

pkg/shell/runner.go
```diff
@@ -65,14 +74,30 @@ func (r *shellRunner) executeCmd(obfuscate func([]string), cmdExe string, cmdArg
...
 }
 
-func (r *shellRunner) execute(cmd *exec.Cmd) error {
+func (r *shellRunner) execute(cmd *exec.Cmd, output io.Writer) error {
 	cmd.Env = append(cmd.Env, os.Environ()...)
 	cmd.Env = append(cmd.Env, r.context.Export()...)
 	cmd.Stdin = os.Stdin
-	cmd.Stdout = os.Stdout
+	if output != nil {
+		cmd.Stdout = io.MultiWriter(os.Stdout, output)
```
northtyphoon: Will it also write image digest query cmd output to os.stdout and eventually go to user's build log? It seems not necessary.

pkg/shell/runner.go
```diff
@@ -65,14 +74,30 @@ func (r *shellRunner) executeCmd(obfuscate func([]string), cmdExe string, cmdArg
 		obfuscate(displayValues)
 	}
 	logrus.Infof("Running command %s %s", cmdExe, strings.Join(displayValues, " "))
-	return r.execute(cmd)
+	var output io.Writer
```
northtyphoon: I think if you change readOuputs from bool to io.Writer, and QueryCmd pass *bytes.Buffer and ExecuteCmd pass nil, the code will be cleaner.

pkg/commands/docker.go
shhsu: I always leave the space when I do the query but I suppose I do not need it.

pkg/grok/docker_compose.go
shhsu: I usually name it buildtimes when I need to iterate. I guess I am not doing it here

pkg/shell/runner.go
shhsu: sure

pkg/shell/runner.go
shhsu: Not sure about this one. I always show command and its outputs and I think we should still show the outputs even if the output is captured by the builder and used.

pkg/build.go
```diff
@@ -51,11 +53,69 @@ type Target interface {
...
+// ImageReference defines the reference to a docker image
+type ImageReference struct {
+	Registry   string `json:"registry"`
+	Repository string `json:"repository"`
+	Tag        string `json:"tag,omitempty"`
+	Digest     string `json:"digest"`
+	reference  reference.Reference
+}
+
+// NewImageReference parse a path of a image and create a ImageReference object
```
ehotinger: NIT: some wording throughout docs. (parse**s**, create**s**). Same comment applies for `String()`

ehotinger:
LGTM. :shipit:

pkg/grok/docker_compose.go
shhsu: Fixed

PR #45, opened 2018-01-22
Update Travis CI to simply call docker build

### LINKED ISSUE #42: Run docker build in Travis CI
Feature Request.

ehotinger:
**Purpose of the PR:**

Updates Travis CI to simply call `docker build`.

**Fixes #42**

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 sudo: required
 
+language: go
+
 services:
   - docker
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: go
 services:
   - docker
 
-before_script:
+before_install:
   - docker build --rm -t acr-builder .
 
 notifications:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,30 +1,12 @@
...
 
-language: go
+services:
+  - docker
 
-go:
-  - 1.9
-  - master
+before_install:
+  - docker build -t acr-builder .
```
shhsu: Yeah, next step is push and we would have to include the kraterdev registry name in the tag. But that's beyond the scope of this change

sajayantony:
@ehotinger - Did you check why travis failed?

ehotinger:
@sajayantony yeah, Travis CI apparently defaults to ruby as the language if one isn't specified and runs `rake` at the end. We have no rake file so naturally it failed. The work around is to either use `language: generic`, but I think I'll just put `language: go` back in in case we use a `Makefile` or whatever down the road. Originally I removed it to (hopefully) reduce some overhead - but it doesn't seem too worthwhile.

PR #47, opened 2018-01-23
Adds dockerignore

### LINKED ISSUE #46: Add a dockerignore file

Feature request. Some unnecessary files can be filtered out by the docker daemon.

ehotinger:
**Purpose of the PR:**

Adds a basic `dockerignore` file so that some unnecessary files can be filtered out by the daemon.

**Fixes #46**

---

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+**/.dockerignore
+**/Dockerfile
+.git
+LICENSE
+Makefile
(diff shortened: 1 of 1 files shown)


ehotinger:
NB: some tests are using actual `Dockerfile`s, so the check to ignore them needs to be less aggressive. Probably just the root one for now.

PR #65, opened 2018-02-15
Update exit codes to be 1 for general all-purpose errors.

### LINKED ISSUE #64: Don't use os.Exit(-1)

Bug report.

**What happened?**:

If the builder fails to run, it will have its exit code be `255` (exit status out of range) if you run `docker inspect` on it. Values should be 0..255 inclusive, where 1 is a general failure. http://tldp.org/LDP/abs/html/exitcodes.html

The container should produce standard exit codes from `docker inspect` based on different types of failures.

Crash the builder by passing in garbage arguments. Run `docker ps -a` and do `docker inspect`

**Fix**
This is fixed by simply changing `os.Exit(-1)` in `main.go` to be 1, but we should define standard exit codes so we can determine which piece failed.

ehotinger:
**Purpose of the PR:**
Updates error codes to be 1 if the builder fails unexpectedly.

**Fixes #64**

---

### DIFF
--- a/main.go
+++ b/main.go
@@ -80,13 +80,13 @@ func main() {
 		if len(os.Args) < 2 {
 			flag.CommandLine.Usage()
 		}
-		os.Exit(-1)
+		os.Exit(1)
 	}
 
 	output, err := json.Marshal(dependencies)
 	if err != nil {
 		logrus.Errorf("Failed to serialize dependencies %s", err)
-		os.Exit(-1)
+		os.Exit(1)
 	}
 	fmt.Printf("\nBuild duration: %s\n", duration)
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -80,13 +80,13 @@ func main() {
 		if len(os.Args) < 2 {
 			flag.CommandLine.Usage()
 		}
-		os.Exit(-1)
+		os.Exit(1)
```
northtyphoon: Can you use a variable named like general_error_code or add a commens so that other people can be aware the conversion?

ehotinger: @northtyphoon + @shhsu - done. Also refactored the other const files to be one-liners rather than duplicating `const` everywhere.

PR #69, opened 2018-02-28
support secret build arg

### LINKED ISSUE #63: Support --docker-secret-build-arg
FEATURE REQUEST
**What happened?**:
Allow passing --docker-secret-build-arg and hide the value from the output or log.

northtyphoon:
**Purpose of the PR:**

**Fixes #63 

#### input
--docker-secret-build-arg sk1=sv1 --docker-secret-build-arg sk2=sv2 --docker-build-arg k1=v1

#### log output
--build-arg k1=v1 --build-arg sk1=************* --build-arg sk2=*************

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Run the following on your project directory to build the project and push to a d
 * `--docker-context-dir` Docker build context. Optional, to be used alongside `--dockerfile`.
 * `--docker-compose-file` Docker Compose file to be invoked for build and push.
 * `--docker-build-arg` Build arguments to be passed to docker build or docker-compose build. This parameter can be specified multiple times.
-* `--docker-secret-build-arg` Build arguments to be passed to docker build or docker-compose build. The argument value contains secret which will be hidden from the log. This parameter can be specified multiple times.
+* `--docker-secret-build-arg` Build arguments to be passed to docker build or docker-compose build. The argument value contains a secret which will be hidden from the log. This parameter can be specified multiple times.
 * `--build-env` Custom environment variables defined for the build process. This parameter can be specified multiple times. (For more details, see `Build Environment`).
 * `--push` Specify if push is required if build is successful.
 * `--verbose` Enable verbose output for debugging.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -41,6 +41,7 @@ Run the following on your project directory to build the project and push to a d
 * `--docker-context-dir` Docker build context. Optional, to be used alongside `--dockerfile`.
 * `--docker-compose-file` Docker Compose file to be invoked for build and push.
 * `--docker-build-arg` Build arguments to be passed to docker build or docker-compose build. This parameter can be specified multiple times.
+* `--docker-secret-build-arg` Build arguments to be passed to docker build or docker-compose build. The argument value contains secret which will be hidden from the log. This parameter can be specified multiple times.
```
ehotinger: Some minor wording: "The argument value contains **a** secret which will be hidden from the log"

pkg/commands/docker.go
```diff
@@ -93,12 +95,34 @@ func (t *dockerBuildTask) Build(runner build.Runner) error {
...
+	}
+
 	if t.contextDir != "" {
 		args = append(args, t.contextDir)
 	} else {
 		args = append(args, ".")
 	}
-	return runner.ExecuteCmd("docker", args)
+
+	return runner.ExecuteCmdWithObfuscation(func(args []string) {
```
ehotinger: Can you make this into an actual method - it's an identical concept for compose.

README.md
ehotinger: Should we put some sort of disclaimer here? Passing secrets through command line isn't really safe anyways since command history, etc. will still be kept on the machine. It's fine in our particular case but for the generic builder it's a bit iffy. I.e. similar to the warning message you see from Docker. This also won't protect people from doing things like `echo $var` in `docker logs` but so it might be a tad misleading.

pkg/commands/docker.go
```diff
@@ -93,12 +95,34 @@ func (t *dockerBuildTask) Build(runner build.Runner) error {
...
+		if len(t.buildSecretArgs) > 0 {
```
ehotinger: Conceptually I'd like to have a substitution map so we can reuse any obfuscation work and lookups in multi step builds later.

README.md
northtyphoon: I can think of reading the value from a file, similar to how docker saves the credential in the config file. If we choose that route, I'd like to do it for all other credentials we pass through the command line.

pkg/commands/docker.go
northtyphoon: Sure. I actually thought about removing all compose codes.

PR #76, opened 2018-03-13
Remove build duration from logs

ehotinger:
**Purpose of the PR:**

This PR completely removes build duration from the builder's output. Instead, the caller should be measuring the total time of the build request since it needs to anyways. I.e., the caller has to measure any overhead incurred while invoking the builder as well as timeout situations.

### DIFF
--- a/main.go
+++ b/main.go
@@ -91,4 +91,5 @@ func main() {
 	}
 
 	fmt.Printf("\nACR Builder discovered the following dependencies:\n%s\n", string(output))
+	fmt.Println("\nBuild complete")
 }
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -91,6 +90,5 @@ func main() {
 		os.Exit(constants.GeneralErrorExitCode)
 	}
 
-	fmt.Printf("\nBuild duration: %dms\n", int64(duration)/int64(time.Millisecond))
```
ehotinger: NB: before merging, we should append a "Build succeeded" line; previously the build duration more-or-less served this purpose.

Issue #80, opened 2018-03-31
Drop docker-compose support


northtyphoon:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

We have discussed it a few times. Since we will evaluate our build configuration story and may look for other options, I suggest we drop docker-compose support and focus on dockerfile at the current stage. It will also help further cleanup and refactor the code base and reduce the image size.

@sajayantony  
@ehotinger

ehotinger:
big 👍 from me - it will make it even harder to get rid of later if people somehow take a dependency on it

sajayantony:
Approved.

northtyphoon:
#81

PR #81, opened 2018-03-31
remove docker-compose support

### LINKED ISSUE #71: Remove --docker-image requirement when building against dockerfile
FEATURE REQUEST

**What happened?**:
Build with `--docker-file` but without `--docker-image`, the command failed complaining "Image name not specified for docker file "

If the users just want to test the dockerfile build without push, it's not necessary to provide the `--docker-image` parameter.

northtyphoon:
**Purpose of the PR:**

**Fixes #71 #80 


pkg/commands/docker.go
```diff
@@ -160,19 +160,22 @@ func PopulateDigests(runner build.Runner, dependencies []build.ImageDependencies
...
+	if reference != nil {
+		refString := reference.String()
+		output, err := runner.QueryCmd("docker", []string{
+			"inspect", "--format", "\"{{json .RepoDigests}}\"", refString,
+		})
+		if err != nil {
+			return err
+		}
+
+		trimCharPredicate := func(c rune) bool {
```
ehotinger: can you refactor this to be real func and pass it in instead

pkg/commands/archive_test.go
```diff
@@ -64,13 +64,13 @@ func testArchiveSource(t *testing.T, tc obtainTestCase) {
...
 	assert.Nil(t, err)
 
-	var projectDockerComposeFile string
+	var projectDockerFile string
 	if tc.targetDir == "" {
-		projectDockerComposeFile = "docker-compose.yml"
+		projectDockerFile = "Dockerfile"
 	} else {
-		projectDockerComposeFile = filepath.Join(tc.targetDir, "docker-compose.yml")
+		projectDockerFile = filepath.Join(tc.targetDir, "Dockerfile")
```
ehotinger: for `filepath.Join` we should use https://github.com/cyphar/filepath-securejoin - it's not really a big deal but worth keeping it as a reminder somewhere

pkg/commands/docker.go
northtyphoon: For now, I actually prefer to keep it as an anonymous in the function scope as it's very specific to the digest case (eg, it trim double quote). I also want to keep it close to the caller to make it easy to find. We can also refactor is later once we need to reuse the logic =)

PR #84, opened 2018-04-01
add --pull and --no-cache option

### LINKED ISSUE #79: Add --pull in docker build command
FEATURE REQUEST

--pull option always attempts to pull a newer version of the image to make sure the local cache is up to date.
https://docs.docker.com/engine/reference/commandline/build/#options

### LINKED ISSUE #82: Allow --no-cache option in docker build command
FEATURE REQUEST

--no-cache allow not using cache when building the image. enforce clean build if needed.

https://docs.docker.com/engine/reference/commandline/build/#options

northtyphoon:
**Purpose of the PR:**

**Fixes #79 **  
**Fixes #82 ** 

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -53,13 +53,15 @@ func (u *dockerUsernamePassword) Authenticate(runner build.Runner) error {
 
 // NewDockerBuild creates a build target with specified docker file and build parameters
 func NewDockerBuild(dockerfile, contextDir string,
-	buildArgs, buildSecretArgs []string, registry, imageName string) build.Target {
+	buildArgs, buildSecretArgs []string, registry, imageName string, pull, noCache bool) build.Target {
 	return &dockerBuildTask{
 		dockerfile:      dockerfile,
 		contextDir:      contextDir,
 		buildArgs:       buildArgs,
 		buildSecretArgs: buildSecretArgs,
 		pushTo:          fmt.Sprintf("%s%s", registry, imageName),
+		pull:            pull,
+		noCache:         noCache,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/commands/docker.go
```diff
@@ -91,7 +93,13 @@ func (t *dockerBuildTask) ScanForDependencies(runner build.Runner) ([]build.Imag
 }
 
 func (t *dockerBuildTask) Build(runner build.Runner) error {
-	args := []string{"build"}
+	// --pull to make sure the local base image cache is up to date.
+	args := []string{"build --pull"}
```
ehotinger: make this a parameter too, just default it to true

northtyphoon: Thought about it. I deliberately hide it to reduce the parameter numbers since we don't need it and requires higher-up changes which is simply defaulted to true. The only case I can think of is users manually copy the base image and they don't have access the registry. I can certainly make the change if you think it may be the case we want to target in the future.

ehotinger: that would make sense if we were writing code just for our code base, but this is open source, it should be a parameter for other people to use it if they want to, just how Docker doesn't make an assumption about it

pkg/commands/docker.go
```diff
@@ -91,7 +95,17 @@ func (t *dockerBuildTask) ScanForDependencies(runner build.Runner) ([]build.Imag
 }
 
 func (t *dockerBuildTask) Build(runner build.Runner) error {
+	// --pull to make sure the local base image cache is up to date.
```
ehotinger: this comment can be moved down, but NBD

Issue #83, opened 2018-03-31
Failed to build if the from command use arg


northtyphoon:
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG REPORT

**What happened?**:
run acr-build using the following dockerfile with --docker-build-arg tag=latest

```
ARG tag
FROM ubuntu:${tag}
```

Build failed:
level=error msg="invalid reference format"

The dependency scan needs to expand arguments in from command.

shhsu:
https://github.com/Azure/acr-builder/pull/86

PR #86, opened 2018-04-04
grok to understand ARG directive and build args

shhsu:
**Purpose of the PR:**

**Fixes #**
https://github.com/Azure/acr-builder/issues/83

**Notes/Details:**

dependency scanning now knows build args

```release-note
```

### DIFF
--- a/pkg/grok/docker_file.go
+++ b/pkg/grok/docker_file.go
@@ -10,41 +10,66 @@ import (
 )
 
 // ResolveDockerfileDependencies given a docker file, resolve its dependencies
-func ResolveDockerfileDependencies(path string) (string, []string, error) {
+func ResolveDockerfileDependencies(buildArgs []string, path string) (string, []string, error) {
 	file, err := os.Open(path)
 	if err != nil {
 		return "", nil, fmt.Errorf("Error opening dockerfile %s, error: %s", path, err)
 	}
 	scanner := bufio.NewScanner(file)
-
+	context, err := parseBuildArgs(buildArgs)
+	if err != nil {
+		return "", nil, err
+	}
 	originLookup := map[string]string{} // given an alias, look up its origin
 	allOrigins := map[string]bool{}     // set of all origins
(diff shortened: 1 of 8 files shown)


pkg/grok/docker_file.go
```diff
@@ -10,41 +10,66 @@ import (
...
+			case "FROM":
+				if len(tokens) < 2 {
 					return "", nil, fmt.Errorf("Unable to understand line %s", line)
 				}
-				alias := tokens[3]
-				originLookup[alias] = origin
-				// Just ignore the rest of the tokens...
-				if len(tokens) > 4 {
-					logrus.Debugf("Ignoring chunks from FROM clause: %v", tokens[4:])
+				var image = os.Expand(tokens[1], func(key string) string {
```
northtyphoon: Cool it looks os.Expand will handle $arg, ${arg} very well?

shhsu: yes, I used it in build.Context too. Just forgot about it at the moment during our offline conversation.

As you can see both $arg and ${arg} are used in our test example.

ehotinger:
btw @northtyphoon @shhsu didn't have a chance to review this, but does it work for `ENV={blah}` too? There's docs on Docker where all `{}` can be used so just making sure we covered them all 👍

shhsu:
As far as I can tell, you cannot use ENV variable on FROM clause.
ARG goes before FROM and ENV goes after.
I tried a few multi-stage example and none of them seem to work. @ehotinger, do you have any example for ENV applying for this use case?

ehotinger:
@shhsu ya original syntax I posted is a bit off, basically just want to make sure this still works after replacement:

```
ENV abc=hello
ENV abc=bye def=$abc
ENV ghi=$abc
```

I'll review the PR later when I have some free time.

shhsu:
Actually, this PR only apply to dependency scanning logic. If a variable is not used in FROM clause we don't really care about it.

PR #88, opened 2018-04-05
skip push if --docker-image is not set

### LINKED ISSUE #87: build failed with --docker-registry but without --docker-image
BUG REPORT

**What happened?**:
run acr-builder with --docker-registry but without --docker-image

Build succeeded

Build failed as level=error msg="invalid reference format"

northtyphoon:
**Purpose of the PR:**

**Fixes #87**

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -54,12 +54,20 @@ func (u *dockerUsernamePassword) Authenticate(runner build.Runner) error {
 // NewDockerBuild creates a build target with specified docker file and build parameters
 func NewDockerBuild(dockerfile, contextDir string,
 	buildArgs, buildSecretArgs []string, registry, imageName string, pull, noCache bool) build.Target {
+
+	var pushTo string
+	// If imageName is empty, skip push.
+	// If registry is empty, it means push to DockerHub.
+	if len(imageName) > 0 {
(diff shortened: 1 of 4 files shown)


pkg/commands/docker.go
```diff
@@ -54,12 +54,20 @@ func (u *dockerUsernamePassword) Authenticate(runner build.Runner) error {
 // NewDockerBuild creates a build target with specified docker file and build parameters
 func NewDockerBuild(dockerfile, contextDir string,
 	buildArgs, buildSecretArgs []string, registry, imageName string, pull, noCache bool) build.Target {
+
+	var pushTo string
+	// If imageName is empty, skip push.
+	// If registry is empty, it means push to DockerHub.
+	if len(imageName) > 0 {
```
ehotinger: `Push(` will throw if `pushTo` is a string so it will fail a push to docker hub

northtyphoon: The case that pushTo is empty is when the imageName and registry are both empty. The caller function has checked the case when push is enabled.

northtyphoon: I think I should explain the error message in the issue is not about push failed. It's because the dependency parser failed to read the pushTo string to extract the repository name.

ehotinger: I think you're good, I reviewed it again and LGTM. Thanks 👍

PR #90, opened 2018-04-05
Understand archive format

### LINKED ISSUE #89: Bad error message on invalid archive format
Bug report

**What happened?**:
If pass a invalid file to --archive. We see this acr-builder blow up like this:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x81384a9]

goroutine 1 [running]:
compress/gzip.(*Reader).Close(0x0, 0x19444440, 0x194000e0)å
	/usr/local/go/src/compress/gzip/gunzip.go:292 +0x19
github.com/Azure/acr-builder/pkg/commands.unTAR.func1(0x0)
	/go/src/github.com/Azure/acr-builder/pkg/commands/archive.go:83 +0x21
github.com/Azure/acr-builder/pkg/commands.unTAR(0xffe25ad2, 0x15, 0xf7ef8600, 0x19781280, 0x83eab80, 0x1958a748)
	/go/src/github.com/Azure/acr-builder/pkg/commands/archive.go:89 +0x3d8
github.com/Azure/acr-builder/pkg/commands.(*ArchiveSource).Obtain(0x195379a0, 0x83ee1e0, 0x1958abd8, 0x0, 0x0)
	/go/src/github.com/Azure/acr-builder/pkg/commands/archive.go:48 +0x197
github.com/Azure/acr-builder/pkg.(Source).Obtain-fm(0x83ee1e0, 0x1958abd8, 0x0, 0x0)
	/go/src/github.com/Azure/acr-builder/pkg/driver/build.go:227 +0x33
github.com/Azure/acr-builder/pkg/workflow.(*Workflow).ScheduleRun.func1(0x83ee1e0, 0x1958abd8, 0x19537a00, 0x0, 0x0)
	/go/src/github.com/Azure/acr-builder/pkg/workflow/workflow.go:70 +0x2b
github.com/Azure/acr-builder/pkg/workflow.(*Workflow).Run(0x19537a00, 0x83ee1e0, 0x1958abd8, 0x0, 0x0)
	/go/src/github.com/Azure/acr-builder/pkg/workflow/workflow.go:49 +0x8b
github.com/Azure/acr-builder/pkg/driver.(*Builder).Run(0x194e5f7c, 0xffe25abb, 0x8, 0x0, 0x0, 0x0, 0x0, 0xffe25b1c, 0x1b, 0xffe25b47, ...)
	/go/src/github.com/Azure/acr-builder/pkg/driver/build.go:63 +0x2fe
main.main()
	/go/src/github.com/Azure/acr-builder/main.go:70 +0xc5f

Build ID: eastus-9 failed because of an internal error after 21.62147022s

A better error message
acr-builder --archive **https://github.com/sajayantony/dockerfiles

shhsu:
**Purpose of the PR:**
See #89 

**Fixes #**  fixes #89

**Notes/Details:**

```release-note
```

### DIFF
--- a/pkg/commands/archive.go
+++ b/pkg/commands/archive.go
@@ -17,7 +17,7 @@ import (
 	"github.com/Azure/acr-builder/pkg/constants"
 )
 
-var maxHeaderSize = 4
+const maxHeaderSize = 4
 
 // NOTE: only support .tar.gz for now
 var supportedArchiveHeaders = map[byte][]byte{
@@ -56,7 +56,11 @@ func (s *ArchiveSource) Obtain(runner build.Runner) error {
 	if err != nil && err != io.EOF {
 		return fmt.Errorf("Failed to peek context header: %s", err)
 	}
-	if !isSupportedArchive(peek) {
+	supported, err := isSupportedArchive(peek)
+	if err != nil {
+		return fmt.Errorf("Failed to read context header: %s", err)
+	}
+	if !supported {
 		return fmt.Errorf("Unexpected file format for %s", s.url)
 	}
 
@@ -157,7 +161,13 @@ func unTAR(baseDir string, r io.Reader) error {
 	}
 }
 
-func isSupportedArchive(source []byte) bool {
+func isSupportedArchive(source []byte) (bool, error) {
+	if len(source) < 1 {
+		return false, fmt.Errorf("Empty header")
+	}
 	header, found := supportedArchiveHeaders[source[0]]
-	return found && bytes.Equal(source[:len(header)], header)
+	if len(source) < len(header) {
+		return false, fmt.Errorf("Format of the file content cannot be determined. File header is corrupted")
+	}
+	return found && bytes.Equal(source[:len(header)], header), nil
 }
--- a/pkg/commands/archive.go
+++ b/pkg/commands/archive.go
@@ -17,16 +17,12 @@ import (
 	"github.com/Azure/acr-builder/pkg/constants"
 )
 
-var maxHeaderSize = 512
-var supportedArchiveHeaders = map[byte][]byte{
-	// Bzip2
-	0x42: {0x42, 0x5A, 0x68},
+var maxHeaderSize = 4
 
+// NOTE: only support .tar.gz for now
+var supportedArchiveHeaders = map[byte][]byte{
 	// Gzip
 	0x1F: {0x1F, 0x8B, 0x08},
-
-	// Xz
-	0xFD: {0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00},
 }
 
 // ArchiveSource defines source in the form of an archive file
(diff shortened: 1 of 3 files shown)


pkg/commands/archive.go
```diff
@@ -41,11 +50,20 @@ func (s *ArchiveSource) Obtain(runner build.Runner) error {
 		}
 	}()
 
+	buf := bufio.NewReader(response.Body)
+	peek, err := buf.Peek(maxHeaderSize)
+	if err != nil && err != io.EOF {
+		return fmt.Errorf("failed to peek context header: %s", err)
```
northtyphoon: >f [](start = 21, length = 1)

upper case

pkg/commands/archive.go
```diff
@@ -15,6 +17,13 @@ import (
 	"github.com/Azure/acr-builder/pkg/constants"
 )
 
+var maxHeaderSize = 512
+var supportedArchiveHeaders = map[byte][]byte{
+	0x42: {0x42, 0x5A, 0x68},                   // Bzip2
+	0x1F: {0x1F, 0x8B, 0x08},                   // Gzip
+	0xFD: {0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00}, //Xz
```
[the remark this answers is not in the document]
shhsu: Where do you mean?

pkg/commands/archive.go
```diff
@@ -15,6 +17,13 @@ import (
 	"github.com/Azure/acr-builder/pkg/constants"
 )
 
+var maxHeaderSize = 512
```
northtyphoon: >512 [](start = 20, length = 3)

I see the max is 6 bytes from the following list

pkg/commands/archive.go
shhsu: Fixed in next iteration

pkg/commands/archive.go
```diff
@@ -137,3 +155,17 @@ func unTAR(baseDir string, r io.Reader) error {
 		}
 	}
 }
+
+func isSupportedArchive(header []byte) bool {
+	if detectCompression(header) {
+		return true
+	}
+	r := tar.NewReader(bytes.NewBuffer(header))
```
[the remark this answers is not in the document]
shhsu: We should because docker does. I will verify. If it doesn't I will add it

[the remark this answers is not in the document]
shhsu: Actually, since we are only currently support .tar.gz, we should just keep the verification as such.
We can add more archive logic later or somehow unify our logic with docker. I think that's beyond this PR and I will just modify my code to detect gz file instead

pkg/commands/archive.go
shhsu: Because of below discussion I have scope this down to 4 bytes

pkg/commands/archive.go
```diff
@@ -137,3 +156,8 @@ func unTAR(baseDir string, r io.Reader) error {
 		}
 	}
 }
+
+func isSupportedArchive(source []byte) bool {
+	header, found := supportedArchiveHeaders[source[0]]
+	return found && bytes.Equal(source[:len(header)], header)
```
northtyphoon: In theory, if len(header) is larger than cap(souce), the code can throw. I guess buf.Peek probably always return a large slice so the code won't hit the issue. It's better to do length check before byte check.

PR #92, opened 2018-04-05
Adds flags for displaying help & binary version info

ehotinger:
**Purpose of the PR:**

Adds flags for displaying help & binary version info

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 
 	"github.com/Azure/acr-builder/pkg/driver"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -61,6 +67,20 @@ func main() {
...
+		flag.PrintDefaults()
+		return
+	}
+
+	if *versionFlag {
+		fmt.Printf(`%s:
+				go version  : %s
+				go compiler : %s
+				platform    : %s/%s
+			   `, os.Args[0], runtime.Version(), runtime.Compiler, runtime.GOOS, runtime.GOARCH)
```
northtyphoon: which one is the binary version?

ehotinger: I actually removed the binary version stuff so the commit message is a bit wrong, have to add in the VERSION stuff like the runner to do that, and maintain a VERSION.txt file

PR #93, opened 2018-04-06
Faster Travis builds - no testing with Docker

ehotinger:
**Purpose of the PR:**

No reason to test with Docker in `RUN`, it slows everything down significantly by using layers. Just run `make` instead.


ehotinger:
2 min and 30 seconds -> 50 sec

PR #95, opened 2018-04-06
Allow support for multiple tags; automatic pushing to multi-tag; ...

### LINKED ISSUE #94: Support multiple tags

Feature request

Support `docker build -t whenry/fedora-jboss:latest -t whenry/fedora-jboss:v2.1 .` -- arbitrary number of `-t`

ehotinger:
**Purpose of the PR:**

Add support for multiple tags; automatic multi-tag pushes; custom tags.
This follows a similar pattern Docker does with build, you can just chain `-t` indefinitely.

Note that the functionality is a bit different in that we don't support `{image}:tag`, since we never really did to begin with. We could change this, but for now we just always use the image name passed in and then append the tags to the same image name, no changing of image names with multi-tag for now.

Example:

```
./acr-builder --docker-registry swankyregistry.azurecr.io --docker-user ehotinger --docker-password youwish --docker-file Dockerfile --docker-context-dir . --push --docker-image acr-builder -t fancytag -t swankytag -t happytag
```

**Fixes #94** 

**Notes/Details:**

The way `docker push` is set up, all layers for all tags will automatically be processed.

PS: you'll still always get `:latest` by default even if you don't have a custom tag.

### DIFF
--- a/main.go
+++ b/main.go
@@ -41,6 +41,7 @@ func main() {
 	// if the program is launched in docker container, use option -v /var/run/docker.sock:/var/run/docker.sock -v ~/.docker:/root/.docker
 	var dockerUser, dockerPW, dockerRegistry string
 	var buildArgs, buildSecretArgs, buildEnvs stringSlice
+	var tags stringSlice
 	var pull, noCache, push, debug bool
 	var buildNumber string
 	flag.StringVar(&buildNumber, constants.ArgNameBuildNumber, "0", fmt.Sprintf("Build number, this argument would set the reserved %s build environment.", constants.ExportsBuildNumber))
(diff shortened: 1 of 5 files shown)


northtyphoon:
umm, i think we support {reposiotry}:tag through --docker-image arg from the very beginning. not sure why you get the impression. 
What is the e2e experience to achieve? Do we also update CLI to ask users to pass both image name and the optional tags? If CLI still follows docker cli pattern, it can become tricky to transfer to the new builder command. Also all the {repository}:tag from CLI must have the same repository while docker doesn't have the limitation.
I'm wondering if the builder can allow multiple docker-image arguments, the overall experience might be better.

shhsu:
I think this feature makes sense. When acr-builder is first written we were thinking of servicing scenario. When a service pull an image during deployment it pulls from an exact version and the tag would likely be a unique one used for that particular deployment.

This feature caters to the "library" scenario. 

microsoft/dotnet: 2.0.6-sdk-2.1.104-stretch == 2.0-sdk-stretch == 2.0.6-sdk-2.1.104 == 2.0-sdk == 2-sdk == sdk == latest

In this case if you have to provide multiple image argument the syntax might seem a bit verbose.

ehotinger:
Abandoning this for now, will chat more offline and send a revised PR with our decisions.

PR #96, opened 2018-04-06
Support multiple custom images

ehotinger:
**Purpose of the PR:**

- Support tagging multiple custom images.
- Support multiple pushes from the builder.
- Include dependencies of all new images and tags in the output.

Example:

```
sudo ./acr-builder --docker-registry myregistry.azurecr.io --docker-user ehotinger --docker-password somepassword --docker-file Dockerfile --docker-context-dir .  -t acr-builder:newtag -t acr-builder2:anothertag -t acr-builder2:footag --push
```

**Fixes #95**

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -53,13 +53,14 @@ func (u *dockerUsernamePassword) Authenticate(runner build.Runner) error {
 
 // NewDockerBuild creates a build target with specified docker file and build parameters
 func NewDockerBuild(dockerfile, contextDir string,
-	buildArgs, buildSecretArgs []string, registry, imageName string, pull, noCache bool) build.Target {
+	buildArgs, buildSecretArgs []string, registry string, imageNames []string, pull, noCache bool) build.Target {
 
-	var pushTo string
+	var pushTo []string
+
 	// If imageName is empty, skip push.
 	// If registry is empty, it means push to DockerHub.
-	if len(imageName) > 0 {
-		pushTo = fmt.Sprintf("%s%s", registry, imageName)
+	for _, imageName := range imageNames {
+		pushTo = append(pushTo, fmt.Sprintf("%s%s", registry, imageName))
 	}
 
 	return &dockerBuildTask{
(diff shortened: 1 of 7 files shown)


pkg/commands/docker.go
```diff
@@ -53,13 +53,14 @@ func (u *dockerUsernamePassword) Authenticate(runner build.Runner) error {
...
+	buildArgs, buildSecretArgs []string, registry string, imageNames []string, pull, noCache bool) build.Target {
+
+	var pushTo []string
 
-	var pushTo string
 	// If imageName is empty, skip push.
 	// If registry is empty, it means push to DockerHub.
-	if len(imageName) > 0 {
-		pushTo = fmt.Sprintf("%s%s", registry, imageName)
+	for _, imageName := range imageNames {
```
northtyphoon: >imageName [](start = 8, length = 9)

Corner case: will it fail?  --docker-registry myregistry.azurecr.io -t ""

pkg/constants/program_args.go
```diff
@@ -32,7 +32,7 @@ const (
 	ArgNameDockerfile = "docker-file"
 
 	// ArgNameDockerImage is the parameter name for docker image (registry url must be excluded from the image name parameter)
-	ArgNameDockerImage = "docker-image"
+	ArgNameDockerImage = "t"
```
northtyphoon: >ArgNameDockerImage [](start = 1, length = 18)

readme needs to be updated

pkg/commands/docker.go
ehotinger: Yep, good catch. I updated `main.go` with a change so that string slices can't be == `""` - it will skip appending them to the array. This also affected build envs (although I didn't check if it would error there), so a double fix!

pkg/constants/program_args.go
ehotinger: Good catch, updated.

ehotinger:
@shhsu waiting for your review before merging :+1:

shhsu:
LGTM

PR #98, opened 2018-04-11
Docker context passtrhough

shhsu:
**Purpose of the PR:**

Unify acr-builder with docker build
acr-builder -context <any source>
would be equal to
docker build <any source>

The problem is if we pass the build argument over, it's hard to figure out where the dockerfile is and dependency scanning wouldn't work.

**Fixes #** 

**Notes/Details:**

```release-note
```


ehotinger:
Dependency scanning is a hard requirement at this point, absolutely can't remove it. Honestly long-term we don't want to invest in a by-pass solution anyways, we just need to write more code. I think it's fine that we just document inconsistencies with Docker. We can limit functionality as long as we know the gaps, e.g., `git@` isn't supported, whatever. People have the expectation we are identical to them - but it's going to be extremely hard to be 100% compatible with them, especially with OCI later on.

main.go
```diff
@@ -48,6 +49,7 @@ func main() {
 	var pull, noCache, push, debug bool
 	var buildNumber string
 	flag.StringVar(&buildNumber, constants.ArgNameBuildNumber, "0", fmt.Sprintf("Build number, this argument would set the reserved %s build environment.", constants.ExportsBuildNumber))
+	flag.StringVar(&dockerContextString, constants.ArgNameDockerContextString, "", "Working directory for the builder.")
```
ehotinger: honestly before I review anything else, I would like to understand why we can't just have one parameter for all of this called `-c` - there should be no `gitURL`, `dockerfile`, `dockerContextString`, just one parameter. That's what I was expecting for this PR

shhsu: Those parameters can be removed. I just didn't include them here. If you want to control your own working directory where the files are checked out to you can still use them. 
However, this is a feature that we can decide to remove. Do we have to do them all in the same PR. I thought I was only trying to fix the scenario for runner. Which it currently only use a single parameter `archive`. We would replace it with `-context`

ehotinger: I guess I don't understand the purpose of the PR, but in the description you put "[unify with Docker build]," this is a step backwards. Absolutely do not add more parameters, it's a complete mess already. Remove and rename them if you need to, this is what this PR reads like right now:

```
flag.StringVar(&dockerContextString, constants.ArgNameDockerContextString, "", "Working directory for the builder.")
flag.StringVar(&workingDir, constants.ArgNameWorkingDir, "", "Working directory for the builder.")
```

shhsu:
@northtyphoon @ehotinger 

PR revamped majorly. I now removed all parameters discussed and added vsts url detection logic

ehotinger:
Doesn't seem like something like this works:

```
$ acr-builder -push --docker-registry ehotinger.azurecr.io -t "eric:foo" -t "eric:latest" -t "eric" -c "https://github.com/northtyphoon/aspnetcore-helloworld.git#master:HelloWorld" -c "Dockerfile"
```

This should fail for context stuff but it should at least get to building steps.

pkg/commands/docker_source.go
```diff
@@ -0,0 +1,225 @@
...
+	if workingDir != "" {
+		s.tracker, err = ChdirWithTracking(runner, workingDir)
+		if err != nil {
+			return err
+		}
+	}
+	if sourceType == dockerSourceGit {
+		sha, queryErr := runner.QueryCmd("git", []string{"rev-parse", "--verify", "HEAD"})
+		if queryErr != nil {
+			fmt.Fprintf(os.Stderr, "Error querying for git head rev: %s, output will not contain git head revision SHA", queryErr)
```
ehotinger: Why not just use `logrus` everywhere, it's a mixed bag right now

pkg/commands/docker_source.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+	}(response)
+	return s.ensureContextFromReader(runner, progReader, workingDir, dockerfile)
+}
+
+func (s *DockerSource) getWithStatusError(url string) (resp *http.Response, err error) {
+	if resp, err = http.Get(url); err != nil {
+		return nil, err
+	}
+	if resp.StatusCode < 400 {
```
ehotinger: Are there any 300s which are valid, if so add a comment?

pkg/constants/etc.go
```diff
@@ -21,14 +18,19 @@ const (
 	// SourceNameLocal is the name of local source
 	SourceNameLocal = "local directory"
 
+	// SourceNamePassThrough is a pass-through build context. We pass the context dir directly to docker and do nothing with the source
+	SourceNamePassThrough = "docker build context"
```
ehotinger: Unused?

pkg/commands/docker_source.go
```diff
@@ -0,0 +1,225 @@
...
+
+// Export exports the source
+func (s *DockerSource) Export() []build.EnvVar {
+	return []build.EnvVar{
+		{Name: constants.ExportsDockerBuildContext, Value: s.context},
+	}
+}
+
+// Remark makes a remark to the dependencies
+func (s *DockerSource) Remark(runner build.Runner, dependencies *build.ImageDependencies) {
```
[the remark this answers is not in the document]
shhsu: I wanted to give a generic name to the function though. Different type of source would populate different fields for the reference object.
In this case, this source object happens to be a git source and it marks the ImageReference wit git header revision.

pkg/constants/etc.go
shhsu: removed

pkg/commands/docker_source.go
shhsu: Anything under 400 are considered non-error. Comment added

pkg/commands/docker_source.go
shhsu: Changed

PR #100, opened 2018-04-14
Better Archiving

ehotinger:
**Purpose of the PR:**

- There are a lot of bugs W.R.T. formats in the current code. Things like `git archive` aren't supported because we expect a very specific format.

For example, this breaks (BTW, it can corrupt your git repo if you aren't careful too!):

```
git clone https://github.com/deis/example-dockerfile-http
cd example-dockerfile-http
git archive --format tar.gz master > master.tar.gz
(upload to azblob)
sudo ./acr-builder --archive "whatever url from the above step" --docker-file "Dockerfile"
```

- Let moby handle the bulk of the untar for us after downloading.
- This gives us support for many different compression types and formats in the future.
- This PR makes it so that the download and untar happen in isolated folders, before it would actually erase the `acr-builder`'s Dockerfile and all other files at the root if you ran it locally with a context of `.`
- Adds a `make cbuild` command to the `Makefile` to `clean` + `build` without running tests.
- For performance, moby has better buffering and pooling mechanisms we can make use of. Also, instead of trying to convert the bytes from the HTTP request and peeking them for a TAR, we just download the entire thing and THEN untar it afterwards.

Performance example with a fully cached build:

**Before**

```
real    0m1.103s
user    0m0.556s
sys     0m0.068s
```

**After**

```
real    0m0.736s
user    0m0.162s
sys     0m0.049s
```

**Some Gotchas:**
- `docker-context-dir` is **NOT** supported right now. It was sorta supported in some scenarios before I think, but it was a hack? In any case, I want to make sure that nobody is expecting this, otherwise we need to add on the extra work.
- the `temp` directory where all unzipping occurs is **not** cleaned up right now. The dep check needs to invoke a clean command. See the TODO in the code for more info.

I.E., this will fail:
```
./acr-builder --archive "https://acrbuild.blob.core.windows.net/public/aspnetcore-helloworld.tar.gz" --docker-file "Dockerfile" --docker-context-dir "Hello"
```

With the error:

```
unable to prepare context: path "Hello" not found
```

Using `Hello/Dockerfile` as `--docker-file` works as expected of course.

This is because the workflow around the archiving makes the assumption that you're at the root of the repo.


pkg/commands/archive.go
```diff
@@ -1,29 +1,23 @@
...
-	// Gzip
-	0x1F: {0x1F, 0x8B, 0x08},
-}
+const (
+	tempWorkingDir = "temp"
+	tempFileName   = "temp.tar.gz"
+)
 
 // ArchiveSource defines source in the form of an archive file
 // Currently we only support tar.gz
```
northtyphoon: Do we still have the limitation?

ehotinger: Na, I just named it that for clarity and if you use software to open it you can use it automatically. The extension doesn't do anything. We can use bzip2 and stuff if we want, I haven't tested all the formats though.

PR #109, opened 2018-04-21
De-dup provided image names to avoid duplicate builds, pushes, and deps

### LINKED ISSUE #108: De-Dup multi-tag image builds
```
sudo ./acr-builder -push --docker-registry ehotinger.azurecr.io -t "eric:foo" -t "eric:latest" -t "eric"
```

This will build `eric:latest` twice and output duplicate dependencies.

ehotinger:
**Purpose of the PR:**

Optimization and correctness. De-duplicate any provided image names to avoid duplicate builds, pushes, and dependency generation. For example, before if you passed in `-t eric -t eric:latest`, we would build twice against `latest`. Same goes if you were to even do `-t foo -t foo -t foo`. It's just a list and not a dictionary, so it's extremely easy to abuse. Now it's a hash and all image names that aren't a digest or have a tag are flagged as `latest`.

**Fixes #108**

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"strings"
 
+	build "github.com/Azure/acr-builder/pkg"
 	"github.com/Azure/acr-builder/pkg/driver"
 	"github.com/Azure/acr-builder/pkg/shell"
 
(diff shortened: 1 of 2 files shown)


sajayantony:
Good catch.

PR #102, opened 2018-04-17
Add library to the fully qualified URI of Docker Hub images

djyou:
**Purpose of the PR:**

The current output of image dependency for a Docker Hub image is invalid. `registry.hub.docker.com/docker` is not a pullable image while `registry.hub.docker.com/library/docker` or just `docker` are:
```
{
    "type": "Runtime",
    "repositoryName": "docker",
    "tag": "17.12.0-ce-git",
    "digest": "sha256:14d3f9d0781478fb34802a197666dd7be5df0ffce46be2bd1dc2f0d814cfdcc9",
    "registryLoginServer": "registry.hub.docker.com"
}
```

We can either change `repositoryName` to `library/docker` or change `registryLoginServer` to empty. I preferred the latter one as the registry `registry.hub.docker.com` plus `library` is much less known to users.

**Notes/Details:**

```release-note
```

@Azure/azure-container-registry

PR #113, opened 2018-04-22
Add library to the fully qualified URI of Docker Hub images

### LINKED ISSUE #112: Add library to the fully qualified URI of Docker Hub images
From #102  - Add library to the fully qualified URI of Docker Hub images.

> The current output of image dependency for a Docker Hub image is invalid. registry.hub.docker.com/docker is not a pullable image while registry.hub.docker.com/library/docker or just docker are:
> 
> {
>     "type": "Runtime",
>     "repositoryName": "docker",
>     "tag": "17.12.0-ce-git",
>     "digest": "sha256:14d3f9d0781478fb34802a197666dd7be5df0ffce46be2bd1dc2f0d814cfdcc9",
>     "registryLoginServer": "registry.hub.docker.com"
> }
> We can either change repositoryName to library/docker or change registryLoginServer to empty. I preferred the latter one as the registry registry.hub.docker.com plus library is much less known to users.

ehotinger:
**Purpose of the PR:**

Adds some additional logic to #102 to fully support the fully qualified URI on Docker Hub images and de-duplicating some `FROM` statements.

**Command**:
```
acr-builder -push -docker-file tests/resources/dockerfile-dups/ReductionDockerfile -t "library/eric:foo" -docker-registry ehotinger.azurecr.io
```

**Example Dockerfile**:

```
FROM alpine
FROM library/alpine
FROM alpine:latest
FROM library/alpine:latest

FROM library/golang
FROM golang
FROM library/golang:latest
FROM golang:latest

FROM ubuntu:latest
FROM ubuntu
FROM library/ubuntu
FROM library/ubuntu:latest
```

**Output:**

```
[
    {
        "image": {
            "registry": "ehotinger.azurecr.io",
            "repository": "library/eric",
            "tag": "foo",
            "digest": "sha256:92c414f2cf966fd4c6d60bc71180737bf878eeca3654cad82991a1e02d885d37"
        },
        "runtime-dependency": {
            "registry": "registry.hub.docker.com",
            "repository": "library/ubuntu",
            "tag": "latest",
            "digest": "sha256:9ee3b83bcaa383e5e3b657f042f4034c92cdd50c03f73166c145c9ceaea9ba7c"
        },
        "buildtime-dependency": [
            {
                "registry": "registry.hub.docker.com",
                "repository": "library/alpine",
                "tag": "latest",
                "digest": "sha256:7b848083f93822dd21b0a2f14a110bd99f6efb4b838d499df6d04a49d0debf8b"
            },
            {
                "registry": "registry.hub.docker.com",
                "repository": "library/golang",
                "tag": "latest",
                "digest": "sha256:024beba2e5a25b36b9dba9e6f0df3cee6c24734a98e77d4f382634e7d9a042c4"
            },
            {
                "registry": "registry.hub.docker.com",
                "repository": "library/ubuntu",
                "tag": "latest",
                "digest": "sha256:9ee3b83bcaa383e5e3b657f042f4034c92cdd50c03f73166c145c9ceaea9ba7c"
            }
        ]
    }
]
```

**Fixes #112**

### DIFF
--- a/pkg/build.go
+++ b/pkg/build.go
@@ -93,7 +93,7 @@ func NewImageReference(path string) (*ImageReference, error) {
 			// DockerHub
 			if result.Registry == "" {
 				result.Registry = DockerHubRegistry
-				result.Repository = reference.Path(named)
+				result.Repository = strings.Join([]string{"library", reference.Path(named)}, "/")
 			} else {
 				// The domain is the DockerHub user name
 				result.Registry = DockerHubRegistry
@@ -149,6 +149,14 @@ func NewImageDependencies(env *BuilderContext, image string, runtime string, bui
 	dict := map[string]bool{}
 	for _, buildtime := range buildtimes {
 		bt := NormalizeImageTag(env.Expand(buildtime))
+
+		// If the image is prefixed with "library/", remove it for comparisons.
+		// "library/" will be added again during image reference generation.
+		// This prevents duplicate dependencies when reading "library/golang" and
+		// "golang" from the Dockerfile.
+		if strings.HasPrefix(bt, "library/") {
+			bt = bt[8:]
+		}
 
 		// If we've already processed the tag after normalization, skip dependency
 		// generation. I.e., they specify "golang" and "golang:latest"
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -241,9 +241,16 @@ func getRepoDigest(jsonContent string, reference *build.ImageReference) string {
 	// Output: <empty>
 
 	prefix := reference.Repository + "@"
+	// If the reference has "library/" prefixed, we have to remove it - otherwise
+	// we'll fail to query the digest, since image names aren't prefixed with "library/"
+	if strings.HasPrefix(prefix, "library/") && reference.Registry == build.DockerHubRegistry {
+		prefix = prefix[8:]
+	}
+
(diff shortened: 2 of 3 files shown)


pkg/build.go
```diff
@@ -150,6 +150,14 @@ func NewImageDependencies(env *BuilderContext, image string, runtime string, bui
 	for _, buildtime := range buildtimes {
 		bt := NormalizeImageTag(env.Expand(buildtime))
 
+		// If the image is prefixed with "library/", remove it for comparisons.
+		// "library/" will be added again during image reference generation.
+		// This prevents duplicate dependencies when reading "library/golang" and
+		// "golang" from the Dockerfile.
+		if strings.HasPrefix(bt, "library/") {
```
northtyphoon: Does it also apply for non docker-hub image? eg,

abc.azurecr.io/library/web
abd.azurecr.io/web

northtyphoon: Never mind, I see the bt is from the original name. 

Another corner case is users may write something like:
```
From registry.hub.docker.com/library/Ubuntu
From Ubuntu
```
But I think dup is OK for such case.

ehotinger: yeah that's a good corner case, I will make a note of that. We need to rewrite these libs eventually so I can use them in the next builder and they're not perfect at normalizing things right now

PR #130, opened 2018-06-12
Enable Windows container builds with acr-builder.

mnltejaswini:
This PR enables windows container builds using acr-builder. As per https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility , Older containers will run the same on newer hosts with hyper-v isolation. Hence added a flag to enable hyper-v isolation for the docker build. Added the corresponding Dockerfile to build acr-builder for windows host.

Note:
We are using a patched version for docker CLI to workaround the named pipe issue. Please refer to https://github.com/Microsoft/go-winio/pull/78 for more details.

### DIFF
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -1,0 +1,87 @@
+FROM microsoft/windowsservercore:1803 as environment
+
+# set the default shell as powershell.
+# $ProgressPreference: https://github.com/PowerShell/PowerShell/issues/2138#issuecomment-251261324 
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
+# install MinGit (especially for "go get" and docker build by git repos) 
+ENV GIT_VERSION 2.17.0
+ENV GIT_TAG v${GIT_VERSION}.windows.1
+ENV GIT_DOWNLOAD_URL https://github.com/git-for-windows/git/releases/download/${GIT_TAG}/MinGit-${GIT_VERSION}-64-bit.zip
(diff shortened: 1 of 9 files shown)


pkg/commands/docker.go
```diff
@@ -136,6 +138,10 @@ func (t *dockerBuildTask) ScanForDependencies(runner build.Runner) ([]build.Imag
 func (t *dockerBuildTask) Build(runner build.Runner) error {
 	args := []string{"build"}
 
+	if t.hypervIsolation {
```
ehotinger: add in an `isolation` parameter instead so we can specify it to whatever we want; e.g. process/namespace isolation is good enough for our use cases, no need for hyper-v partition isolation

pkg/commands/docker.go
```diff
@@ -54,7 +54,7 @@ func (u *dockerUsernamePassword) Authenticate(runner build.Runner) error {
 
 // NewDockerBuild creates a build target with specified docker file and build parameters
 func NewDockerBuild(dockerfile string,
-	buildArgs, buildSecretArgs []string, registry string, imageNames []string, pull, noCache bool) build.Target {
+	buildArgs, buildSecretArgs []string, registry string, imageNames []string, hypervIsolation, pull, noCache bool) build.Target {
```
ehotinger: tests will have to be updated with the new param

Dockerfile.Windows
```diff
@@ -0,0 +1,87 @@
+FROM microsoft/windowsservercore:1803 as environment
```
northtyphoon: Can you make the base image as ARG?

PR #153, opened 2018-07-06
Simplifies templates

### LINKED ISSUE #138: Simplify template rendering
- Templates can be simplified to just base/values at this point. Remove any need for a directory and specific layout.

ehotinger:
**Purpose of the PR:**

- Simplifies templates, essentially removing the notion of a "Job," where a "Job" could have multiple templates (child templates, partials, etc.). Now, the rendering engine simply takes a template along with a values ("Config") file.
- Adds in support for overriding values in `exec` with `--set`

**Fixes #138**


cli/exec.go
```diff
@@ -58,7 +58,9 @@ func newExecCmd(out io.Writer) *cobra.Command {
 
 	f := cmd.Flags()
 
-	f.StringVar(&e.stepsFile, "steps", "", "the steps file to use when building")
+	f.StringVar(&e.valuesFile, "values", "", "the values file to use")
+	f.StringVar(&e.stepsFile, "steps", "", "the steps file to use")
```
ehotinger: maybe we can change this to `--template`?

PR #154, opened 2018-07-07
Adds the render command, useful for linting templates.

### LINKED ISSUE #140: Template linting
- Implement `acb lint` to verify whether or not a template will render successfully.

ehotinger:
**Purpose of the PR:**
- Adds the render command, useful for linting templates.
- Refactors the base render options and related flags to be shared between commands, e.g. `exec` and `render`.
- Allows values from files to be completely optional, i.e. you can `exec` just a pipeline without any `{{.Values}}` required from an associated `values.toml`. NB: this change also allows you to just use `--set` instead, and still reference them by `{{.Values.blah}}`, so it's much more flexible. 
- Updates tests to include `{{.Build.Registry}}` as a reserved keyword.

**Fixes #140**

### DIFF
--- a/templating/base_render_options.go
+++ b/templating/base_render_options.go
@@ -64,7 +64,7 @@ func OverrideValuesWithBuildInfo(c1 *Config, c2 *Config, options *BaseRenderOpti
 	return base, nil
 }
 
-// LoadAndRenderSteps loads a steps file and renders it according to an optional values file, --set values,
+// LoadAndRenderSteps loads a template file and renders it according to an optional values file, --set values,
 // and base render options.
 func LoadAndRenderSteps(opts *BaseRenderOptions) (string, error) {
 	template, err := LoadTemplate(opts.StepsFile)
@@ -94,7 +94,7 @@ func LoadAndRenderSteps(opts *BaseRenderOptions) (string, error) {
 		return "", fmt.Errorf("Failed to override values: %v", err)
 	}
 
-	engine := New()
+	engine := NewEngine()
 	rendered, err := engine.Render(template, mergedVals)
 	if err != nil {
 		return "", fmt.Errorf("Error while rendering templates: %v", err)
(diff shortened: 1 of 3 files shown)


templating/base_render_options.go
```diff
@@ -43,3 +63,55 @@ func OverrideValuesWithBuildInfo(c1 *Config, c2 *Config, options BaseRenderOptio
...
+
+		setConfig = &Config{RawValue: rawVals, Values: map[string]*Value{}}
+	}
+
+	mergedVals, err := OverrideValuesWithBuildInfo(config, setConfig, opts)
+	if err != nil {
+		return "", fmt.Errorf("Failed to override values: %v", err)
+	}
+
+	engine := New()
```
northtyphoon: Rename to NewEngine?  I was a little confused at first glance =)

templating/base_render_options.go
```diff
@@ -43,3 +63,55 @@ func OverrideValuesWithBuildInfo(c1 *Config, c2 *Config, options BaseRenderOptio
 	base["Values"] = vals
 	return base, nil
 }
+
+// LoadAndRenderSteps loads a steps file and renders it according to an optional values file, --set values,
```
northtyphoon: a template file

templating/base_render_options.go
```diff
@@ -43,3 +63,55 @@ func OverrideValuesWithBuildInfo(c1 *Config, c2 *Config, options BaseRenderOptio
...
+	ret := Values{}
+	for _, v := range values {
+		s := strings.Split(v, "=")
+		if len(s) != 2 {
+			return "", fmt.Errorf("failed to parse --set data: %s", v)
+		}
+		ret[s[0]] = s[1]
+	}
+
+	return ret.ToTOMLString()
```
northtyphoon: How does toml handle the value string containing quote like "value1 value2"?

PR #156, opened 2018-07-07
Update Makefile for Windows

ehotinger:
**Purpose of the PR:**

- Updates the Makefile for Windows, so that `acb.exe` is produced from `make build` via `BINARY_SUFFIX`.
- Loads different `Makefile.{GOOS}` by trying to read `${GOOS}` if Go is installed, with a fallback if Go isn't installed.
- Fixes a bug in `make help`, where the correct text wouldn't be displayed on Windows.
- Adds `make vendor` which runs `dep ensure` - and ran it to update packages again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,34 @@ CTIMEVAR=-X $(PKG)/version.GITCOMMIT=$(GITCOMMIT) -X $(PKG)/version.VERSION=$(VE
 GO_LDFLAGS=-ldflags "-w $(CTIMEVAR)"
 GO_LDFLAGS_STATIC=-ldflags "-w $(CTIMEVAR) -extldflags -static"
 
+ifneq "$(strip $(shell command -v go 2>/dev/null))" ""
+	GOOS ?= $(shell go env GOOS)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -23,6 +23,34 @@ CTIMEVAR=-X $(PKG)/version.GITCOMMIT=$(GITCOMMIT) -X $(PKG)/version.VERSION=$(VE
...
+			endif
+		endif
+	else
+		GOOS ?= $$GOOS
+		GOARCH ?= $$GOARCH
+	endif
+endif
+
+# Try to include OS specific Makefiles, without any warnings/errors if they don't exist.
+-include Makefile.$(GOOS)
```
northtyphoon: >Makefile.$(GOOS) [](start = 9, length = 16)

I see the Makefile.windows in Makefile.windows folder. Do you need to update the path?

ehotinger: Maybe a bug in Codeflow? `Makefile.windows` is just a file, and this loads that file relative to the Makefile being ran.

PR #162, opened 2018-07-09
First class .Build.Date

### LINKED ISSUE #161: Add {{.Build.Date}}
From #159 

> .Build.Date would = 20180709-131422z which means (1:14:22pm UTC)

ehotinger:
**Purpose of the PR:**

First classes `{{.Build.Date}}`, which defaults to `time.Now().UTC()` with `yyyyMMdd-HHmmssz` format.

**Fixes #161**

/cc @SteveLasker

### DIFF
--- a/cli/shared.go
+++ b/cli/shared.go
@@ -4,6 +4,8 @@
 package cli
 
 import (
+	"time"
+
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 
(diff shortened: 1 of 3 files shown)


sajayantony:
Once this is rolled out, I think we need to enable this format to be specified.

ehotinger:
@sajayantony At that point you should just do `{{ now() | format "blah" }}` which is already supported.

PR #165, opened 2018-07-10
Add push to pipelines

### LINKED ISSUE #135: First-class multi-push
- `push` should be a first-class construct in a `.toml`
- normalize image names and prefix them with the registry appropriately
- iterate through all image names after all steps complete and try to execute a push (with retries), log a warning if any image exists in `push` but not locally

ehotinger:
**Purpose of the PR:**
- Add `push` as a first-class construct in a `pipeline`.
- Normalize image names and prefix them with the registry appropriately

**Closes #135**


cli/build.go
```diff
@@ -145,8 +144,7 @@ func (b *buildCmd) run(cmd *cobra.Command, args []string) error {
 	// TODO: create secrets
 	secrets := []*graph.Secret{}
 
-	p := graph.NewPipeline(steps, push, secrets)
-	dag, err := graph.NewDagFromPipeline(p)
+	p, err := graph.NewPipeline(steps, push, secrets, b.opts.Registry, b.registryUser, b.registryPw)
```
northtyphoon: >p [](start = 1, length = 1)

rename to pipeline? I often need to look up to find what p means if I see p is passed as parameter

graph/pipeline.go
```diff
@@ -96,7 +146,57 @@ func (p *Pipeline) initialize() {
 			s.CompletedChan = make(chan bool)
 		}
 
+		// Adjust the run command so that the ACR registry is prefixed for all tags
+		s.Run = util.PrefixTags(s.Run, p.RegistryName)
```
northtyphoon: >PrefixTags [](start = 15, length = 10)

PrefixTags might be able to return Tags directly

cli/build.go
ehotinger: renamed all instances

graph/pipeline.go
ehotinger: can't here, because the step's run command will be used directly during build; we have to modify the run command so that that tag names get passed correctly to the scanner + docker

in reality I just added `Tags` and `BuildArgs` to make it easier/"cleaner" to read the node processing logic

cli/build.go
ehotinger: Er... looks like my `push` didn't go through because GH is having status up issues... and I merged already, I'll combine it with the next PR.

graph/pipeline.go
northtyphoon: I see. I thought PrefixTags could return both the prefixed Run string and prefixed Tags array so you don't need to call ParseTags again. I notice ParseTags also does additional work to dedup.

PR #174, opened 2018-07-16
Convert TOML to YAML

ehotinger:
**Purpose of the PR:**

- Converts all usage of TOML to YAML.


templating/config.go
```diff
@@ -5,8 +5,8 @@ package templating
 
 // Config represents configuration values.
 type Config struct {
-	RawValue string            `json:"rawValue,omitempty"`
```
northtyphoon: I assume you never used the json tag and just clean it up here?

ehotinger: Correct

PR #177, opened 2018-07-19
Fix the warnings about loading config and setting up credential helper.

### LINKED ISSUE #176: WARNING regarding config file  when doing Windows container builds
Windows container build shows the following warnings

- [x] Fix - WARNING: Error loading config file: C:\Users\ContainerAdministrator\.docker\config.json: invalid character 'ÿ' looking for beginning of value
WARNING! Your password will be stored unencrypted in C:\Users\ContainerAdministrator\.docker\config.json.

- [ ] Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

mnltejaswini:
Setup Windows credential helper and fix the docker config.json file warnings.

Fixes #176

### DIFF
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -66,6 +66,26 @@ RUN $url = ('https://golang.org/dl/go{0}.windows-amd64.zip' -f $env:GOLANG_VERSI
 	\
 	Write-Host 'Complete.';
 
+# Setup Windows Credential helper.
+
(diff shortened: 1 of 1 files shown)


mnltejaswini:
@ehotinger  Yes. I will confirm and make changes to the new acr-builder. That will be tracked as a different item.

PR #179, opened 2018-07-26
Allow specifying home volume name

### LINKED ISSUE #178: Build failed if ARG value is quoted in the Dockerfile
BUG REPORT

**What happened?**:
Build failed with the error msg="invalid reference format"

```
ARG buildimagetag="2.1-sdk"
ARG runtimeimagetag="2.1-aspnetcore-runtime"
FROM microsoft/dotnet:${buildimagetag} AS build
COPY /code /code
WORKDIR /code
RUN dotnet build -c Release -o output
FROM microsoft/dotnet:${runtimeimagetag}
COPY --from=build /code/output .
EXPOSE 80
ENTRYPOINT ["dotnet","code.dll"]
ENV ASPNETCORE_URLS http://*:80
```

https://github.com/Azure/acr/issues/115

ehotinger:
**Purpose of the PR:**

- Add download command to `scanner`
- Move some files to `pkg` so they can be consumed later
- Add `acb.yaml` (linux + windows)
- Simplify image names
- Adds support for "bring your own" volumes, wherein you can specify the home volume name and the builder will just reuse the pre-existing volume

**Fixes #178**


baseimages/scanner/scan/dependencies.go
```diff
@@ -238,5 +238,13 @@ func parseAssignment(in string) (string, string, error) {
...
-	return values[0], values[1], nil
+
+	val := removeSurroundingQuotes(values[1])
+
+	return values[0], val, nil
+}
+
+// removeSurroundingQuotes removes all surrounding quotes.
+func removeSurroundingQuotes(s string) string {
+	return strings.Trim(s, `"`)
```
northtyphoon: >`"` [](start = 24, length = 3)

Do you need to handle single quote?

northtyphoon:
> 	return strings.Join(out, ", ")

nit:Build the output directly, something like strings.Builder?

---
Refers to: pkg/util/errors.go:25 in 50c2e16. [](commit_id = 50c2e1649de123d0a41ccf5a99daf200358735e0, deletion_comment = False)

baseimages/scanner/scan/dependencies.go
ehotinger: Yes, good catch. Fixed.

PR #189, opened 2018-08-03
Aggregate image dependencies across multiple build steps

### LINKED ISSUE #184: Dependency Scanning Output
- [ ] Only output results of dependency scanning once, at the end of all steps.
- [ ] Aggregate all dependencies into a single JSON.

ehotinger:
**Purpose of the PR:**

- Aggregate all image dependencies across multiple `build` steps and flush this information once at the end of the entire pipeline. 
- Simplify DAG traversal by first-classing the root node; `Nodes` will consist of all other vertexes, making a `range` loop easier.
- Update dockerignore to add a few additional files
- When querying digest, use `buf` instead of `os.StdErr`, this is a bug.
- Remove unused mutex from `builder`
- Misc. small refactorings
- Add a test for `getRepoDigest`
- Update README
- Use `log` instead of `fmt` in several places to add timing information in critical sections.
- Ignore errors in several places which are unrecoverable

**Fixes #184**

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -53,18 +53,14 @@ func (b *Builder) RunAllBuildSteps(ctx context.Context, pipeline *graph.Pipeline
 		}
 	}
 
-	root := pipeline.Dag.Nodes[graph.RootNodeID]
 	var completedChans []chan bool
 	errorChan := make(chan error)
-	for k, v := range pipeline.Dag.Nodes {
-		if k == graph.RootNodeID {
-			continue
-		}
-		completedChans = append(completedChans, v.Value.CompletedChan)
+	for _, n := range pipeline.Dag.Nodes {
+		completedChans = append(completedChans, n.Value.CompletedChan)
 	}
 
-	for _, n := range root.Children() {
-		go b.processVertex(ctx, pipeline, root, n, errorChan)
+	for _, n := range pipeline.Dag.Root.Children() {
+		go b.processVertex(ctx, pipeline, pipeline.Dag.Root, n, errorChan)
 	}
 
 	// Block until either:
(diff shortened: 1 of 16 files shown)


builder/builder.go
```diff
@@ -8,16 +8,16 @@ import (
...
 
 	"github.com/Azure/acr-builder/baseimages/scanner/models"
-	"github.com/Azure/acr-builder/baseimages/scanner/util"
+	scannerutil "github.com/Azure/acr-builder/baseimages/scanner/util"
 	"github.com/Azure/acr-builder/graph"
 	"github.com/Azure/acr-builder/pkg/taskmanager"
-	builderutil "github.com/Azure/acr-builder/util"
+	"github.com/Azure/acr-builder/util"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
```
northtyphoon: >"github.com/sirupsen/logrus" [](start = 1, length = 28)

Can it be replaced by log/

builder/builder.go
```diff
@@ -90,48 +82,39 @@ func (b *Builder) RunAllBuildSteps(ctx context.Context, pipeline *graph.Pipeline
...
 	}
 
-	for k, v := range pipeline.Dag.Nodes {
-		if k == graph.RootNodeID {
+	var deps []*models.ImageDependencies
+	for _, n := range pipeline.Dag.Nodes {
+		step := n.Value
+		log.Printf("Step ID %v marked as %v (elapsed time in seconds: %f)\n", step.ID, step.StepStatus, step.EndTime.Sub(step.StartTime).Seconds())
+
+		if !step.IsBuildStep() { // If the step isn't a build step, it won't have dependencies from the scanner.
```
northtyphoon: >IsBuildStep [](start = 11, length = 11)

Is it possible to just check ImageDependencies is nil or empty so you don't need to special handle build?

builder/builder.go
ehotinger: Sure, I went ahead and replaced it.

builder/builder.go
ehotinger: Fixed

PR #205, opened 2018-08-10
Don't render the content if deserialization fails

ehotinger:
**Purpose of the PR:**

Don't render the content if deserialization fails, simply output the error and a message to run `acb lint` locally.


templating/values.go
```diff
@@ -69,7 +69,7 @@ func merge(c *Config, merged map[string]interface{}) (map[string]interface{}, er
 
 	vals, err := Deserialize([]byte(c.RawValue))
 	if err != nil {
-		return merged, fmt.Errorf("Failed to deserialize values during merge: %s, Err: %v", c.RawValue, err)
+		return merged, fmt.Errorf("Failed to deserialize values. Try linting your template locally. Err: %v", err)
```
ehotinger: @sajayantony better wording? Most people won't know `acb lint`, if I say `try linting your template locally with Azure Container Builder`, it will be hard to find this project since it's branded as acr-builder and not acb or AzureContainerBuilder.

PR #207, opened 2018-08-10
Better parsing for --set

ehotinger:
**Purpose of the PR:**

Better parsing for `--set`, allow values to contain equal signs in them.

### DIFF
--- a/templating/base_render_options.go
+++ b/templating/base_render_options.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 // BaseRenderOptions represents additional information for the composition of the final rendering.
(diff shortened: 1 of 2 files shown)


templating/base_render_options.go
```diff
@@ -112,14 +114,22 @@ func LoadAndRenderSteps(template *Template, opts *BaseRenderOptions) (string, er
...
-		s := strings.Split(v, "=")
-		if len(s) != 2 {
-			return "", fmt.Errorf("failed to parse --set data: %s", v)
+		i := strings.Index(v, "=")
+		if i < 0 {
+			return "", errors.New("failed to parse --set data; invalid format, no = assignment found")
+		}
+		key := v[:i]
+		if key == "" {
+			return "", errors.New("failed to parse --set data; invalid format, empty key")
```
northtyphoon: empty key => better to tell what is the expected format, eg "--set key=value"?

ehotinger: SGTM

PR #213, opened 2018-08-13
First-class build in Step

### LINKED ISSUE #200: First-class build
Instead of `run: build ...`, use `build: ...`

ehotinger:
**Purpose of the PR:**

First-classes `build` as a property of `Step`, see the changed yaml for usage references.

**Fixes #200**

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -160,7 +160,7 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step) error {
 	var args []string
 
 	if step.IsBuildStep() {
-		dockerfile, context := parseDockerBuildCmd(step.Cmd)
+		dockerfile, context := parseDockerBuildCmd(step.Build)
 		volName := b.workspaceDir
 
 		deps, err := b.scrapeDependencies(ctx, volName, step.WorkDir, step.ID, dockerfile, context, step.Tags, step.BuildArgs)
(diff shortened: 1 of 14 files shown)


builder/builder.go
```diff
@@ -182,12 +182,12 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step) error {
...
 
-			step.Cmd = replacePositionalContext(step.Cmd, ".")
+			step.Build = replacePositionalContext(step.Build, ".")
 		}
 
 		args = b.getDockerRunArgs(volName, workDir, step)
-		args = append(args, "docker")
-		args = append(args, strings.Fields(step.Cmd)...)
+		args = append(args, "docker", "build")
+		args = append(args, strings.Fields(step.Build)...)
```
northtyphoon: >(args, strings.Fields(step.Build) [](start = 15, length = 33)

Are users already be able to to pass arbitrary arguments to build and run commands?

northtyphoon:
> 			args = append(args, "--env", env)

--env --entrypoint don't apply to build, is it correct?

---
Refers to: builder/builder.go:194 in 1a38e87. [](commit_id = 1a38e87c0c452bdee5753e26dc5fd81a1a925cb7, deletion_comment = False)

builder/builder.go
ehotinger: yes

ehotinger:
@northtyphoon correct, those don't apply to `build`

PR #214, opened 2018-08-13
Best-effort csv-style parsing for cmd, add back in args

### LINKED ISSUE #198: Task does not recognize && when running the container.

  execute yaml with 
```
steps: 
  - run: bash -c "echo FROM hello-world > dockerfile && ls"
  - run: bash ls 
```

**What happened?**:
```
2018/08/10 01:51:49 Successfully set up Docker configuration
2018/08/10 01:51:49 Step args: [docker run --name acb_step_0 --volume /Users/sajay/code/src/github.com/AzureCR/acr-task/examples:/workspace --volume /var/run/docker.sock:/var/run/docker.sock --volume home:/acb/home --env HOME=/acb/home --workdir /workspace bash -c "echo FROM hello-world > dockerfile && ls"]
FROM: -c: line 0: unexpected EOF while looking for matching `"'
FROM: -c: line 1: syntax error: unexpected end of file
Error: failed to run step id: acb_step_0: exit status 1
```
When running docker the dockerfile gets created 
```
docker run --rm -it bash -c 'echo FROM hello-world > dockerfile && ls'
bin         etc         media       root        srv         usr
dev         home        mnt         run         sys         var
dockerfile  lib         proc        sbin        tmp
```

ehotinger:
**Purpose of the PR:**

- Adds back in the `args` field. `cmd` takes precedence over `args` if both are present. Double quotes are the only thing we support, there is no reason to use single quotes as a delimiter.

Examples:

```
steps:
  - cmd: bash -c "echo FROM busybox\nCOPY /hello RUN cat /hello > third.txt"
  - cmd: bash -c "echo hello-world > test.txt && ls && cat test.txt"
  - args:
    - 'bash'
    - '-c'
    - |
        echo "FROM busybox
        COPY /hello 
        RUN cat /hello" > someDockerfile
```

**Fixes #198**

### DIFF
--- a/graph/step.go
+++ b/graph/step.go
@@ -4,8 +4,10 @@
 package graph
 
 import (
+	"encoding/csv"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/Azure/acr-builder/baseimages/scanner/models"
@@ -19,13 +21,14 @@ const (
 
 var (
 	errMissingID    = errors.New("Step is missing an ID")
-	errMissingProps = errors.New("Step is missing a cmd or build property")
+	errMissingProps = errors.New("Step is missing a cmd, build, or args property")
 )
 
 // Step is a step in the execution task.
 type Step struct {
 	ID            string   `yaml:"id"`
 	Cmd           string   `yaml:"cmd"`
+	Args          []string `yaml:"args"`
 	Build         string   `yaml:"build"`
 	WorkDir       string   `yaml:"workDir"`
 	EntryPoint    string   `yaml:"entryPoint"`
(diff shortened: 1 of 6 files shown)


graph/step.go
```diff
@@ -125,3 +127,21 @@ func (s *Step) HasNoWhen() bool {
...
+func (s *Step) GetArgs() []string {
+	if len(s.Args) <= 0 {
+		s.Args = toArgv(s.Cmd)
+	}
+	return s.Args
+}
+
+func toArgv(s string) []string {
+	r := csv.NewReader(strings.NewReader(s))
+	r.Comma = ' '
```
northtyphoon: >Comma [](start = 3, length = 5)

do we allow tab?

ehotinger: no, only single space

PR #218, opened 2018-08-14
Invoke /bin/sh by default to use the shell's field splitting algorithm

ehotinger:
**Purpose of the PR:**

Invoke /bin/sh by default to use the shell's field splitting algorithm

**Example**

```
steps:
  - cmd: |
      bash -c 'echo "FROM golang:1.10-alpine AS gobuild-base
      RUN apk add --no-cache \
        git \
        make

      FROM gobuild-base AS base
      WORKDIR /go/src/github.com/ehotinger/scratch
      COPY . .
      RUN make static && mv scratch /usr/bin/scratch

      FROM scratch
      COPY --from=base /usr/bin/scratch /usr/bin/scratch
      ENTRYPOINT [ "scratch" ]
      CMD [ ]" > test.Dockerfile && cat test.Dockerfile'
  - cmd: 'bash ls'
  - build: -f test.Dockerfile -t scratch .
  - cmd: 'bash rm test.Dockerfile'
  - cmd: 'bash ls'
```

### DIFF
--- a/builder/context.go
+++ b/builder/context.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"path"
 	"regexp"
+	"runtime"
 	"strings"
 
 	"github.com/Azure/acr-builder/graph"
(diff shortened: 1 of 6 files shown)


builder/context.go
```diff
@@ -28,40 +29,53 @@ var (
...
+	step *graph.Step,
+	envs []string,
+	entrypoint string,
+	cmd string) []string {
+
+	var args []string
+	// Run user commands from a shell instance in order to mirror the shell's field splitting algorithms,
+	// so we don't have to write our own argv parser for exec.Command.
+	if runtime.GOOS == "windows" {
+		// TODO: finalize on how to handle the Windows shell.
```
northtyphoon: likely need to enclose the command in {}

ehotinger: will make a separate issue for windows support

PR #234, opened 2018-08-17
Adds timeouts for implicit build steps and adds some additional log messages

### LINKED ISSUE #230: Add timeouts for implicit steps
There are several implicit steps as a part of `build`. Not all of these steps have an associated `stepTimeout` context tied to them, i.e. `login`, `scan`, etc. All of these implicit steps must have their own pre-defined timeout, otherwise we end up in a situation like https://github.com/Azure/acr/issues/122 where a bug can result in implicit contexts consuming the maximum execution time.

### LINKED ISSUE #231: Additional logs for all implicit steps
Not all logs can be shared when executing `scan` or later, self-invoking `acb`, but there are still places where logs are missing, i.e. when scraping dependencies, digests, etc. All implicit steps should be logged to ease debugging pains.

ehotinger:
**Purpose of the PR:**

- Adds timeouts for implicit build steps and adds some additional log messages

**Fixes #231**
**Fixes #230**

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -41,13 +41,19 @@ func NewBuilder(pm *procmanager.ProcManager, debug bool, workspaceDir string) *B
 func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 	if !b.procManager.DryRun {
 		log.Printf("Setting up Docker configuration...")
-		if err := b.setupConfig(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


ehotinger:
@northtyphoon I made the timeouts rather large, so let me know if you think we should cut them down in a few places. It would be pretty absurd to hit any of these timeouts except for `scrape`. Since `scrape` does download + parse, it should be kinda big.

PR #237, opened 2018-08-19
avoid "operation not permitted"  in unpacking tar

### LINKED ISSUE #233: Untar fails if there's a .git folder included
Untar seems to fail during `build` if a `.git` folder is included.

See https://github.com/Azure/acr/issues/122

northtyphoon:
**Purpose of the PR:**

Ref: #233 

@ehotinger, I didn't find a way to overwrite the owner through TarOption. Since I think we only use the docker archive lib in the particular untar scenario, the change updated the archive.go directly. If you already starts to integrate other solution, feel free to close the PR.


ehotinger:
@northtyphoon we can merge and test this but please leave #233 open since this isn't a pure fix. We'll need to spend more time making a proper fix.

Issue #243, opened 2018-08-21
Add tests for Build CLI


ehotinger:
- Add tests for build CLI; validate that the rendered template and so forth are correctly generated. See #242 
- Don't scan for tags twice; return tags after adding prefixes and set push tags equal to the returned values

PR #244, opened 2018-08-21
Prefix registry name to image name for build command

### LINKED ISSUE #242: acb build doesn't prefix image tag
BUG REPORT

**What happened?**:

`acb build --registry myregistry.azurecr.io --push -t image:v1 ...`

It failed as the image was tagged as `image:v1` but pushed to `myregistry.azurecr.io/image:v1`

The output image is tagged as `myregistry.azurecr.io/image:v1` and pushed to the registry successfully.

It looks the prefix only applies on the push not on the `rendered` build step.

https://github.com/Azure/acr-builder/blob/04dfc488769578ae2ef403d0e9e2c3e65c20ea70/cmd/build.go#L120-L141

ehotinger:
**Purpose of the PR:**

- Prefix registry name to image name for build command

**Fixes #242**

### DIFF
--- a/util/prefixes.go
+++ b/util/prefixes.go
@@ -17,3 +17,17 @@ func PrefixRegistryToImageName(registry string, img string) string {
 
 	return img
 }
+
+// PrefixTags prefixes tags in the specified command and returns the new command.
+func PrefixTags(cmd string, registry string) string {
+	if registry == "" {
+		return cmd
+	}
+	fields := strings.Fields(cmd)
+	for i := 1; i < len(fields); i++ {
+		if fields[i-1] == "-t" || fields[i-1] == "--tag" {
+			fields[i] = PrefixRegistryToImageName(registry, fields[i])
+		}
+	}
+	return strings.Join(fields, " ")
+}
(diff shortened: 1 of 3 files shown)


util/prefixes.go
```diff
@@ -17,3 +17,17 @@ func PrefixRegistryToImageName(registry string, img string) string {
...
+
+// PrefixTags prefixes tags in the specified command and returns the new command.
+func PrefixTags(cmd string, registry string) string {
+	if registry == "" {
+		return cmd
+	}
+	fields := strings.Fields(cmd)
+	for i := 1; i < len(fields); i++ {
+		if fields[i-1] == "-t" || fields[i-1] == "--tag" {
+			fields[i] = PrefixRegistryToImageName(registry, fields[i])
```
northtyphoon: You might be able to return tags array from this function as well so you don't need to parse rendered string twice.

ehotinger: Yeah, definitely. I'm gonna add this to #243 and merge this for now so that you're unblock at least.

PR #262, opened 2018-08-23
use ASCII encoding in config.json

Jinhuafei:
**Purpose of the PR:**
make sure the generated config.json file is ascii encoding.

**Fixes #**


Jinhuafei:
The change should fix https://github.com/Azure/acr-builder/issues/176

PR #263, opened 2018-08-24
Minor patches

### LINKED ISSUE #143: First-class networks
- [x] Support networks that can be created at the start of a pipeline and shared between steps
- [x] Add `ports`,  `-p` for publishing ports
- [x] Allow `--name` to be specified via `id`
- [x] Allow specifying `--network` to interact with the host

### LINKED ISSUE #241: Not accounting for comments
**BUG REPORT**:

**What happened?**:
**task.yaml**
```sh
version: 1.0.0
steps:
  - run: bash ls
  - run: build -t helloworld:release .
  - run: git clone https://github.com/demo42/deploy.git
  - run: bash ls 
    workDir: ./deploy
  #- run: {{.Run.Registry}}/functions/helm:latest 
```
` az acr build -f tasks.yaml .`
errored:
```sh
Error: Error while rendering templates: Failed to execute template: tasks.yaml. Err: template: tasks.yaml:8:16: executing "tasks.yaml" at <.Run.Registry>: can't evaluate field Registry in type interface {}
```

**Environment**:

### LINKED ISSUE #252: Allow Home volume name to be accessible through Run
- Allow the home volume's name to be accesible through `.Run.`

### LINKED ISSUE #254: Rename step property envs to env

### LINKED ISSUE #258: Step: continue on error
Introduce a bool that allows a Step to be marked as successful despite having an error.

### LINKED ISSUE #259: Consider moving push as a step type
The more acb is about running containers I believe we feel that push should just move to a step type 

```
steps:
- build: -t 
- push: Image , image, image...
```

### LINKED ISSUE #260: Parser: ignore comments
Stuff like `# FROM blah` might cause issues. Best to just skip the lines altogether.

Trivial impl:

```
if strings.TrimSpace(newline) == ""  || strings.HasPrefix(strings.TrimSpace(newline), "#") { 
    continue
}
```

### LINKED ISSUE #265: Rename network to networks
Networks should be plural and support passing in multiple `--network` commands. As a part of #143

ehotinger:
**Purpose of the PR:**

Minor patches for various issues. Each commit corresponds to a particular issue. For the easiest review, I suggest reviewing each commit separately as they're completely isolated.

**Fixes #265**
**Fixes #260**
**Fixes #259**
**Fixes #258**
**Fixes #254**
**Fixes #252**
**Fixes #241**
**Fixes #143**

### DIFF
--- a/builder/context.go
+++ b/builder/context.go
@@ -61,8 +61,8 @@ func (b *Builder) getDockerRunArgs(
 	if step.User != "" {
 		sb.WriteString(" --user " + step.User)
 	}
-	for _, network := range step.Networks {
-		sb.WriteString(" --network " + network)
+	if step.Network != "" {
+		sb.WriteString(" --network " + step.Network)
 	}
 	if step.Isolation != "" {
 		sb.WriteString(" --isolation " + step.Isolation)
--- a/builder/context.go
+++ b/builder/context.go
@@ -61,8 +61,8 @@ func (b *Builder) getDockerRunArgs(
 	if step.User != "" {
 		sb.WriteString(" --user " + step.User)
 	}
-	if step.Network != "" {
-		sb.WriteString(" --network " + step.Network)
+	for _, network := range step.Networks {
+		sb.WriteString(" --network " + network)
 	}
 	if step.Isolation != "" {
 		sb.WriteString(" --isolation " + step.Isolation)
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -201,7 +201,7 @@ func (b *buildCmd) createBuildTask() (*graph.Task, error) {
 	// TODO: create secrets
 	secrets := []*graph.Secret{}
 
-	return graph.NewTask(steps, secrets, b.opts.Registry, b.registryUser, b.registryPw)
+	return graph.NewTask(steps, secrets, b.opts.Registry, b.registryUser, b.registryPw, true)
 }
 
 func (b *buildCmd) createRunCmd() string {
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -112,10 +112,10 @@ func (b *buildCmd) run(cmd *cobra.Command, args []string) error {
 	ctx := context.Background()
 
 	procManager := procmanager.NewProcManager(b.dryRun)
-	if b.opts.SharedContextDir == "" {
+	if b.opts.SharedContextDirectory == "" {
 		if !b.dryRun {
 			homeVolName := fmt.Sprintf("%s%s", volume.VolumePrefix, uuid.New())
-			b.opts.SharedContextDir = homeVolName
+			b.opts.SharedContextDirectory = homeVolName
 			v := volume.NewVolume(homeVolName, procManager)
 			if msg, err := v.Create(ctx); err != nil {
 				return fmt.Errorf("Err creating docker vol. Msg: %s, Err: %v", msg, err)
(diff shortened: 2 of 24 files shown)


graph/testdata/acb.yaml
```diff
@@ -15,7 +15,7 @@ steps:
   - id: puller
     cmd: azure/images/docker pull ubuntu
     entryPoint: someEntryPoint
-    envs: [eric=foo, foo=bar]
+    env: [eric=foo, foo=bar]
     secretEnvs: [someAkvSecretEnv]
```
northtyphoon: should the format of secretEnvs be same as  env?

docs/acb.md
```diff
@@ -154,4 +156,8 @@ The `build` property of a step specifies how to build a set of images. If build
 
 ### startDelay
 
-`startDelay` can be used to delay a step's execution. This is an integer value measured in seconds.
\ No newline at end of file
+`startDelay` can be used to delay a step's execution. This is an integer value measured in seconds.
+
+### ignoreErrors
+
+If `ignoreErrors` is set to `true`, the step will be marked as complete regardless of whether or not an error occurred during its execution.
```
northtyphoon: it's better to clarify the default value for optional settings.

templating/template.go
```diff
@@ -24,3 +28,21 @@ func (t *Template) GetData() []byte {
...
+	for _, line := range lines {
+		tLine := strings.TrimSpace(line)
+		if !strings.HasPrefix(tLine, "#") {
+			// Append the original line to preserve any spacing.
+			ret = append(ret, line)
+		}
+	}
+	return &Template{
+		Name: name,
+		Data: []byte(strings.Join(ret, "\n")),
```
northtyphoon: Is the ending "\n" required?

builder/context.go
```diff
@@ -61,8 +61,8 @@ func (b *Builder) getDockerRunArgs(
 	if step.User != "" {
 		sb.WriteString(" --user " + step.User)
 	}
-	if step.Network != "" {
-		sb.WriteString(" --network " + step.Network)
+	for _, network := range step.Networks {
```
northtyphoon: Does `docker run` allow multiple networks? How does it work?

cmd/build.go
```diff
@@ -42,7 +42,6 @@ type buildCmd struct {
 	isolation       string
 	network         string
 	platform        string
-	homeVol         string
```
northtyphoon: What was the use case of the homeVol before? I see the code now always creates a new homeVol as SharedContextDir. I'm just curious if some scenarios won't work.

builder/context.go
ehotinger: No, but it's possible to connect multiple. I'm actually going to revert this change for now for simplicity, we can decide later if we want to introduce it

docs/acb.md
ehotinger: Updated.

templating/template.go
ehotinger: Not sure I understand - we split by `\n` so we're just adding them back in at this point.

graph/testdata/acb.yaml
ehotinger: `secretEnvs` should probably be deleted for now until MSI work is done.

templating/template.go
northtyphoon: Oh, I see.

PR #279, opened 2018-09-08
Some changes for acb on windows

Jinhuafei:
1. disable network creation
2. run/build with hyperv isolation by default
3. read setupconfig image name from env


builder/builder_windows.go
```diff
@@ -0,0 +1,320 @@
...
+	args := []string{"docker", "rm", "-f"}
+	for _, n := range task.Dag.Nodes {
+		step := n.Value
+		if step.StepStatus != graph.Skipped {
+			killArgs := append(args, step.ID)
+			_ = b.procManager.Run(ctx, killArgs, nil, nil, nil, "")
+		}
+	}
+
+	// for _, network := range task.Networks {
```
ehotinger: Delete this file; rename the other one back to what it was; go back to the previous one and add `if runtime.GOOS != "windows"` or make a helper file: `network_windows.go` and invoke this there, where `_windows` would just be a `return` no-op.

builder/setup_windows.go
```diff
@@ -18,6 +19,11 @@ const (
 
 // setupConfig initializes ~/.docker/config.json
 func (b *Builder) setupConfig(ctx context.Context) error {
+	imageName := ""
+	if imageName = os.Getenv("ACB_CONFIGIMAGENAME"); imageName == "" {
```
ehotinger: can you lift this to its own func; `imageName := getConfigImageName()`

builder/builder_windows.go
```diff
@@ -0,0 +1,320 @@
...
+		procManager:  pm,
+		debug:        debug,
+		workspaceDir: workspaceDir,
+	}
+}
+
+// RunTask executes a Task.
+func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
+	if !b.procManager.DryRun {
+		// for _, network := range task.Networks {
```
ehotinger: Same comment here

builder/builder_windows.go
```diff
@@ -0,0 +1,320 @@
...
+			if util.IsGitURL(dockerContext) || util.IsVstsGitURL(dockerContext) {
+				workingDirectory = step.ID + "/" + getContextFromGitURL(dockerContext)
+			} else {
+				workingDirectory = step.ID
+			}
+
+			step.Build = replacePositionalContext(step.Build, ".")
+		}
+		// Needs hyperv isolation for older version windows images
+		args = b.getDockerRunArgs(volName, workingDirectory, step, nil, "", "docker build --isolation hyperv "+step.Build)
```
ehotinger: `isolation` is already provided from `build.go` - don't default it here. If people build with Windows in a task, we should default it during step creation as a more sane default, otherwise just leave it off.

graph/step.go
```diff
@@ -141,3 +143,9 @@ func (s *Step) IsBuildStep() bool {
 func (s *Step) IsPushStep() bool {
 	return len(s.Push) > 0
 }
+
+func (s *Step) UpdateBuildStepWithDefaults() {
+	if s.IsBuildStep() && runtime.GOOS == "windows" && !strings.Contains(s.Build, "--isolation") {
+		s.Build = fmt.Sprintf("--isolation hyperv %s", s.Build)
```
ehotinger: shouldn't this be `%s --isolation hyperv`?

pkg/procmanager/procmanager.go
```diff
@@ -53,6 +53,7 @@ func (pm *ProcManager) Run(
 	cmd.Stdout = stdOut
 	cmd.Stderr = stdErr
 
+
```
ehotinger: nit: remove the line

graph/step.go
Jinhuafei: Doesn't the docker build command end with context? s.Build doesn't include "docker build". 

docker build [OPTIONS] PATH | URL | -

ehotinger: Oh right, so it doesn't matter here actually.

PR #290, opened 2018-09-12
Allow GOOS and GOARCH to be accessible from Run

ehotinger:
**Purpose of the PR:**

- Allow GOOS and GOARCH to be accessible from Run templating.

### DIFF
--- a/templating/values_test.go
+++ b/templating/values_test.go
@@ -122,7 +122,7 @@ func TestOverrideValuesWithBuildInfo(t *testing.T) {
 	expectedGitTag := "some git tag"
 	expectedSharedContextDirectory := "acb_home_vol_12345"
 	expectedOS := "linux"
-	expectedArch := "amd64"
+	// expectedArch := "amd64" // TODO: Not exposed yet.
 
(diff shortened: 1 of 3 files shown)


templating/values_test.go
```diff
@@ -139,6 +141,8 @@ func TestOverrideValuesWithBuildInfo(t *testing.T) {
 		GitTag:      expectedGitTag,
 		Date:        parsedTime,
 		SharedContextDirectory: expectedSharedContextDirectory,
+		OS:   expectedOS,
+		Arch: expectedArch,
```
ehotinger: @SteveLasker  do you prefer `.Run.Arch` or `.Run.Architecture`

templating/values_test.go
```diff
@@ -157,6 +161,8 @@ func TestOverrideValuesWithBuildInfo(t *testing.T) {
 		{"{{.Run.GitTag}}", expectedGitTag},
 		{"{{.Run.Date}}", expectedTime},
 		{"{{.Run.SharedContextDirectory}}", expectedSharedContextDirectory},
```
northtyphoon: As we discussed, need to revise the value if it's a volume name.

PR #300, opened 2018-09-14
Print out the network being used during step execution

### LINKED ISSUE #296: Print out step network
This requires some refactoring of the Windows code; this code needs to be deleted:

```
if runtime.GOOS == "windows" {
		cmd = append(cmd, "--driver")
		cmd = append(cmd, "nat")
	}
```

And `Network` needs to have a `Driver` string property instead.

ehotinger:
**Purpose of the PR:**

- Explicitly declare the driver as `bridge` on Linux and `nat` on Windows.
- Add the `Driver` property to `Network`.
- Print out the network being used during step execution
- Print out the network's driver type during network creation.
- Add unit tests for network

**Fixes #296**

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -39,7 +39,7 @@ func NewBuilder(pm *procmanager.ProcManager, debug bool, workspaceDir string) *B
 // RunTask executes a Task.
 func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 	for _, network := range task.Networks {
-		log.Printf("Creating Docker network: %s\n", network.Name)
+		log.Printf("Creating Docker network: %s, driver: '%s'\n", network.Name, network.Driver)
 		if msg, err := network.Create(ctx, b.procManager); err != nil {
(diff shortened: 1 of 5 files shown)


builder/builder.go
```diff
@@ -39,7 +39,7 @@ func NewBuilder(pm *procmanager.ProcManager, debug bool, workspaceDir string) *B
 // RunTask executes a Task.
 func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 	for _, network := range task.Networks {
-		log.Printf("Creating Docker network: %s\n", network.Name)
+		log.Printf("Creating Docker network: %s, driver: '%s'\n", network.Name, network.Driver)
```
northtyphoon: how about `"%+v", network`?

ehotinger: Gonna leave as-is for now, we can add a `ToString` if we want to print out the full network contents later at some point. I'm not even sure how much value there is in printing out the driver.

PR #339, opened 2018-10-26
set total timeout for build task

### LINKED ISSUE #338: acb build timeout is 1 hour
BUG
**What happened?**:
build step timeout is 8 hours but the total build task timeout is 1 hour.

northtyphoon:
**Purpose of the PR:**

**Fixes #338**


graph/task.go
```diff
@@ -155,18 +144,6 @@ func (t *Task) initialize() error {
 		t.TotalTimeout = t.StepTimeout
 	}
 
-	if t.StepTimeout < minStepTimeoutInSeconds {
```
sajayantony: curious why we are allowing anything to set a minTimeout of a value less than 10 seconds.

ehotinger: step/task timeout isn't provided in most task-based yamls out of the box so we need this default initialization logic. What's the intent behind deleting this block?

cmd/build.go
```diff
@@ -194,7 +194,7 @@ func (b *buildCmd) createBuildTask() (*graph.Task, error) {
 	// TODO: create secrets
 	secrets := []*graph.Secret{}
 
-	return graph.NewTask(steps, secrets, b.opts.Registry, b.registryUser, b.registryPw, true)
+	return graph.NewTask(steps, secrets, b.opts.Registry, b.registryUser, b.registryPw, buildTimeoutInSec+pushTimeoutInSec, true)
```
ehotinger: I guess the constant should be login + build + push + buffer, maybe we can just define it in its own constant.

graph/task.go
ehotinger: same, not sure the intent of removing the min/max stuff. But the rest LGTM.

ehotinger: I think we still want these min/max bounds. Rest LGTM.

cmd/build.go
ehotinger: btw I made https://github.com/Azure/acr-builder/issues/340 as a separate item, don't have to tackle it in this PR but should make sure it doesn't regress

graph/task.go
```diff
@@ -19,21 +19,9 @@ const (
 	// The default step timeout is 10 minutes.
 	defaultStepTimeoutInSeconds = 60 * 10
 
-	// The minimum step timeout is 10 seconds.
-	minStepTimeoutInSeconds = 10
-
-	// The maximum step timeout is 2 hours.
-	maxStepTimeoutInSeconds = 60 * 60 * 2
```
ehotinger: this should be increased to 8h

graph/task.go
northtyphoon: We have several cases which have different timeout settings. The services also has timeout control. It's tricky to sync all of them. It becomes hard to reason which timeout is hit when uses report "context deadline exceeded" error.
My suggestion is 
1) Keep acb logic simple, less default 
2) Give users full control, don't override it unless it truly impacts the functionality and it's hard for users to figure out. For example, if users set negative timeout value, we should override it or report error. If users want to timeout 5 seconds on a particular step, I don't see why we need to override it.

PR #347, opened 2018-11-30
use default isolation to run build step in windows

### LINKED ISSUE #346: Windows build: build step should use default isolation to avoid performance hit
BUG

**What happened?**:
For windows, all steps are run by hyperv isolation by default (https://github.com/Azure/acr-builder/blob/474e36f751f87dd1cd44c67f21527c47747c34f9/builder/context.go#L44) However it comes with a cost. The following is the comparison.

docker run docker build -h => 6-8 seconds
docker run --isolation=hyperv docker build -h  => 40-60 seconds

To avoid the performance hit, the docker-cli image (`docker`) in windows should be always compatible with the host os and the build step itself is run by default isolation.

northtyphoon:
**Purpose of the PR:**

**Fixes #346 **

### DIFF
--- a/builder/context.go
+++ b/builder/context.go
@@ -40,7 +40,9 @@ func (b *Builder) getDockerRunArgs(
 	// so we don't have to write our own argv parser for exec.Command.
 	if runtime.GOOS == "windows" {
 		args = []string{"powershell.exe", "-Command"}
-		if step.Isolation == "" {
+		if step.Isolation == "" && !step.IsBuildStep() {
+			// Use hyperv isolation for non-build steps.
+			// Use default isolation for build step to improve performance. It assumes the docker-cli image is compatbile with the host os.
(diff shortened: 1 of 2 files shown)


builder/context.go
```diff
@@ -40,7 +40,9 @@ func (b *Builder) getDockerRunArgs(
 	// so we don't have to write our own argv parser for exec.Command.
 	if runtime.GOOS == "windows" {
 		args = []string{"powershell.exe", "-Command"}
-		if step.Isolation == "" {
+		if step.Isolation == "" && !step.IsBuildStep() {
+			// Use hyperv isolation for non-build steps.
+			// Use default isolation for build step to improve performance. It assumes the docker-cli image is compatbile with the host os.
```
ehotinger: nit: typo here on compatibility

PR #352, opened 2018-12-03
Add a warning message to github contexts that don't contain .git

### LINKED ISSUE #321: Tasks with a git context, without .git, can't find Dockerfiles
**IBUG REPORT**:
Tasks defined with a context intended for github, that doesn't have .git at the end is a valid task creation, but upon execution, says it can't find the dockerfile.
**What happened?**:

```
az acr task create \
  -n helloworld \
  -t helloworld:{{.Run.ID}} \
  -t helloworld:latest \
  -f Dockerfile \
  --arg REGISTRY_NAME=$REGISTRY_NAME/ \
  --context https://github.com/demo42/helloworld \
  --git-access-token $PAT
az acr task run -n helloworld
```
produced an error:
```
2018/09/23 14:20:28 Obtaining source code and scanning for dependencies...
2018/09/23 14:20:29 Output from dependency scanning: Error: Error opening dockerfile /workspace/build/Dockerfile, error: open /workspace/build/Dockerfile: no such file or directory
Error: failed to run step id: build: failed to scan dependencies: exit status 1
```
The problem is it's not clear what the error is. I just know this pattern.

While I know we need to support non-.git extensions, when we know it's github, can we give a better error, or a hint to the problem. Or, if we know it's github, can we fail the task creation?

**Environment**:

ehotinger:
**Purpose of the PR:**

- Add a warning message to github contexts that don't contain .git

**Fixes #321**

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -326,6 +326,7 @@ func getRepoDigest(jsonContent string, reference *image.Reference) string {
 }
 
 func validateDockerContext(context string) {
+	context = strings.ToLower(context)
 	if strings.Contains(context, "github") && !strings.Contains(context, ".git") {
 		log.Printf("WARNING: %s might not be valid context. Valid Git repositories should end with .git.\n", context)
 	}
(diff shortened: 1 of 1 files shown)


ehotinger:
@rosanch This doesn't cover all the cases for docker context, that's not really possible since someone could have a private git server. Plus, there's bitbucket, gitlab, etc. I just put in GitHub for now since it's the most common one. We don't use `logrus` for colors right now just because it's another dependency. We could use it, I'm not opposed to refactoring it all. It's just important to stick to one thing since it'll cause output interleaving.

PR #355, opened 2018-12-03
Adds a test for loading and rendering steps

### LINKED ISSUE #314: Add additional templating tests
Add tests for:

- [ ] `LoadAndRenderSteps` (`base_render_options.go`)
- [ ] `mergeMaps` (`values.go`)

ehotinger:
**Purpose of the PR:**

- Adds a test for loading and rendering steps

**Fixes #314**

### DIFF
--- a/templating/base_render_options_test.go
+++ b/templating/base_render_options_test.go
@@ -3,7 +3,9 @@
 
 package templating
 
-import "testing"
+import (
+	"testing"
+)
 
 func TestParseValues_Valid(t *testing.T) {
 	tests := []struct {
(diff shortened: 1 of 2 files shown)


templating/base_render_options_test.go
```diff
@@ -3,7 +3,9 @@
 
 package templating
 
-import "testing"
+import (
```
rosanch: I've been wondering about this for a while?
What is the difference between the 2? I tried to look for it, but I only see a pattern of this being used when you import more than one file.

ehotinger: There's no difference. It's an autocorrect pattern from `gofmt` starting in Go 1.11

PR #359, opened 2018-12-04
Allows steps to be repeated

### LINKED ISSUE #320: Consider: Add repeat to step
Similar to `retry`, allow a step to re-run itself multiple times.
NB: can be accomplished using `for` template loops already.

ehotinger:
**Purpose of the PR:**
- Allows containers to be repeated after their initial execution. If multiple errors are encountered during the repetition, the last error will be recorded. If the step is marked as `ignoreErrors`, the final error will be ignored and the step will be marked successful.

- Fixes a bug where if a container fails to execute it will have its `retryDelay` invoked. Now we only sleep before the final failure.

**Fixes #320**

### DIFF
--- a/graph/step.go
+++ b/graph/step.go
@@ -51,10 +51,13 @@ type Step struct {
 	Network                         string   `yaml:"network"`
 	Isolation                       string   `yaml:"isolation"`
 	IgnoreErrors                    bool     `yaml:"ignoreErrors"`
-	Retries                         int      `yaml:"retries"`
 	RetryDelayInSeconds             int      `yaml:"retryDelay"`
 	DisableWorkingDirectoryOverride bool     `yaml:"disableWorkingDirectoryOverride"`
 	Pull                            bool     `yaml:"pull"`
+	// Retries specifies how many times a Step will be retried if it fails after its initial execution.
+	Retries int `yaml:"retries"`
+	// Repeat specifies how many times a Step will be repeated after its initial execution.
+	Repeat int `yaml:"repeat"`
 
 	StartTime  time.Time
 	EndTime    time.Time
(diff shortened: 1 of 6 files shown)


graph/step.go
```diff
@@ -51,10 +51,13 @@ type Step struct {
 	Network                         string   `yaml:"network"`
 	Isolation                       string   `yaml:"isolation"`
 	IgnoreErrors                    bool     `yaml:"ignoreErrors"`
-	Retries                         int      `yaml:"retries"`
 	RetryDelayInSeconds             int      `yaml:"retryDelay"`
 	DisableWorkingDirectoryOverride bool     `yaml:"disableWorkingDirectoryOverride"`
 	Pull                            bool     `yaml:"pull"`
+	// Retries specifies how many times a Step will be retried if it fails after its initial execution.
+	Retries int `yaml:"retries"`
```
northtyphoon: I assume the code checks retries and repeats are >0 ?

pkg/procmanager/procmanager.go
```diff
@@ -31,6 +31,30 @@ func NewProcManager(dryRun bool) *ProcManager {
...
+	stdOut io.Writer,
+	stdErr io.Writer,
+	cmdDir string,
+	retries int,
+	retryDelay int,
+	containerName string,
+	repeat int,
+	ignoreErrors bool) error {
+	var err error
+	for i := 0; i <= repeat; i++ {
```
northtyphoon: does repeat include current iteration?

pkg/procmanager/procmanager.go
```diff
@@ -31,6 +31,30 @@ func NewProcManager(dryRun bool) *ProcManager {
...
+		innerErr := pm.RunWithRetries(ctx, args, stdIn, stdOut, stdErr, cmdDir, retries, retryDelay, containerName)
+		if innerErr != nil {
+			err = innerErr
```
northtyphoon: wrap the err from each iteration and return the wrapped one

graph/step.go
ehotinger: Added in validation. It'll throw an error now.

pkg/procmanager/procmanager.go
ehotinger: No

pkg/procmanager/procmanager.go
ehotinger: Switched it to use our internal `errors[]` which has a `String()` method that does `string.Join(",")` on the aggregated list.

PR #362, opened 2018-12-05
add initial rety logic with exponential backoff

shahzzzam:
**Purpose of the PR:**

Add retry logic when a GET request is made to get the context. It makes 3 attempts with some exponential delay

**Fixes #**

### DIFF
--- a/scan/context.go
+++ b/scan/context.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"time"
 
 	"github.com/Azure/acr-builder/util"
 	dockerbuild "github.com/docker/cli/cli/command/image/build"
@@ -24,6 +25,7 @@ import (
 
 const (
 	archiveHeaderSize = 512
+	maxRetries        = 3
 )
 
 // ObtainSourceCode obtains the source code from the specified context.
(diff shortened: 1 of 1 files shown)


scan/context.go
```diff
@@ -112,19 +114,40 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
 
 // getWithStatusError does an http.Get() and returns an error if the
 // status code is 4xx or 5xx.
+// It retries 10 times either if
```
ehotinger: Comment can just be made generic: `it retries if either:`. Comments that are displaced from the variable name are likely to end up out of date

scan/context.go
```diff
@@ -112,19 +114,40 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
...
-		return resp, nil
-	}
-	msg := fmt.Sprintf("failed to GET %s with status %s", url, resp.Status)
-	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return nil, errors.Wrapf(err, msg+": error reading body")
+	attempt := 0
+
+	for attempt < maxRetries {
+		fmt.Printf("Attempt # %d for url %s", attempt, url)
```
ehotinger: Remove the log, no reason to log anything unless it failed.

PR #367, opened 2018-12-06
support multiple registry logins using additional creds

### LINKED ISSUE #361: build/exec cmd: support multiple registries
FEATURE REQUEST

`acb build`  and `acb exec` support login to multiple registries. For example:

`acb build --registry '{"server": "abc.azurecr.io", "username": "abc", "password": "abc"} --registry '{"server": "xyz.azurecr.io", "username": "xyz", "password": "xyz"}'`

shahzzzam:
**Purpose of the PR:**

Should support something like:

    ./acb build . --credentials="reg1;usr1;pw1" --credentials="reg2;usr2;pw2"

`make test` works fine. 

But haven't tested it e2e yet

**Fixes #361**

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -59,16 +59,16 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 	}
 	log.Printf("Successfully set up Docker configuration")
 	if task.UsingRegistryCreds() {
-		log.Printf("Logging in to registry: %s\n", task.RegistryName)
 		timeout := time.Duration(loginTimeoutInSec) * time.Second
 		loginCtx, cancel := context.WithTimeout(ctx, timeout)
 		defer cancel()
 		for _, creds := range task.Credentials {
+			log.Printf("Logging in to registry: %s\n", creds.RegistryName)
 			if err := b.dockerLoginWithRetries(loginCtx, creds.RegistryName, creds.RegistryUsername, creds.RegistryPassword, 0); err != nil {
 				return err
 			}
+			log.Printf("Successfully logged into %s\n", creds.RegistryName)
 		}
-		log.Println("Successfully logged in")
 	}
 
 	var completedChans []chan bool
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -60,9 +60,9 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 	log.Printf("Successfully set up Docker configuration")
 	if task.UsingRegistryCreds() {
 		timeout := time.Duration(loginTimeoutInSec) * time.Second
-		loginCtx, cancel := context.WithTimeout(ctx, timeout)
-		defer cancel()
 		for _, creds := range task.Credentials {
+			loginCtx, cancel := context.WithTimeout(ctx, timeout)
+			defer cancel()
 			log.Printf("Logging in to registry: %s\n", creds.RegistryName)
 			if err := b.dockerLoginWithRetries(loginCtx, creds.RegistryName, creds.RegistryUsername, creds.RegistryPassword, 0); err != nil {
 				return err
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -63,8 +63,10 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 		timeout := time.Duration(loginTimeoutInSec) * time.Second
 		loginCtx, cancel := context.WithTimeout(ctx, timeout)
 		defer cancel()
-		if err := b.dockerLoginWithRetries(loginCtx, task.RegistryName, task.RegistryUsername, task.RegistryPassword, 0); err != nil {
-			return err
+		for _, creds := range task.Credentials {
+			if err := b.dockerLoginWithRetries(loginCtx, creds.RegistryName, creds.RegistryUsername, creds.RegistryPassword, 0); err != nil {
+				return err
+			}
 		}
 		log.Println("Successfully logged in")
 	}
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -38,6 +38,7 @@ type buildCmd struct {
 	target          string
 	registryUser    string
 	registryPw      string
+	credentials     []string
 	isolation       string
 	platform        string
 	tags            []string
@@ -83,6 +84,7 @@ func newBuildCmd(out io.Writer) *cobra.Command {
 
 	f.StringVarP(&r.registryUser, "username", "u", "", "the username to use when logging into the registry")
 	f.StringVarP(&r.registryPw, "password", "p", "", "the password to use when logging into the registry")
+	f.StringArrayVar(&r.credentials, "credentials", []string{}, "all credentials for private repos")
 
 	f.StringVar(&r.isolation, "isolation", "", "the isolation to use")
 	f.StringVar(&r.target, "target", "", "specify a stage to build")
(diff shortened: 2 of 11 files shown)


builder/builder.go
```diff
@@ -63,8 +63,10 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 		timeout := time.Duration(loginTimeoutInSec) * time.Second
 		loginCtx, cancel := context.WithTimeout(ctx, timeout)
 		defer cancel()
-		if err := b.dockerLoginWithRetries(loginCtx, task.RegistryName, task.RegistryUsername, task.RegistryPassword, 0); err != nil {
-			return err
+		for _, creds := range task.Credentials {
```
sajayantony: Do you want to login in parallel (max of 3 in parallel) just to make sure that the overall execution time doesn't increase? Or do you want to optimize that as another item /cc @northtyphoon

cmd/build.go
```diff
@@ -193,7 +195,26 @@ func (b *buildCmd) createBuildTask() (*graph.Task, error) {
...
+	credentials = append(credentials, &graph.Credential{
+		RegistryName:     b.opts.Registry,
+		RegistryUsername: b.registryUser,
+		RegistryPassword: b.registryPw,
+	})
+
+	// Add any additional creds provided by the user in the --credentials flag
+	for _, creds := range b.credentials {
+		// creds should be of the form of "regName;userName;password"
+		splitCreds := strings.Split(creds, ";")
```
ehotinger: Make a function for this and add unit tests. One of the tests should be `foo;bar;somepas;swordwith;semicolons` -> `[]{foo, bar, somepas;swordwith;semicols}` - will not work atm.

cmd/build.go
```diff
@@ -193,7 +195,26 @@ func (b *buildCmd) createBuildTask() (*graph.Task, error) {
 	// TODO: create secrets
 	secrets := []*graph.Secret{}
 
-	return graph.NewTask(steps, secrets, b.opts.Registry, b.registryUser, b.registryPw, taskTotalTimeoutInSec, true)
+	var credentials []*graph.Credential
```
ehotinger: nit: you can pre-allocate the array size rather than calling `append` multiple times

cmd/build.go
```diff
@@ -83,6 +84,7 @@ func newBuildCmd(out io.Writer) *cobra.Command {
 
 	f.StringVarP(&r.registryUser, "username", "u", "", "the username to use when logging into the registry")
 	f.StringVarP(&r.registryPw, "password", "p", "", "the password to use when logging into the registry")
+	f.StringArrayVar(&r.credentials, "credentials", []string{}, "all credentials for private repos")
```
ehotinger: `run` needs this as well; make a few functions to share between them for the split and append

graph/task.go
```diff
@@ -197,16 +194,22 @@ func (t *Task) initialize() error {
 
 // UsingRegistryCreds determines whether or not the Task is using registry creds.
 func (t *Task) UsingRegistryCreds() bool {
-	return t.RegistryName != "" &&
-		t.RegistryPassword != "" &&
-		t.RegistryUsername != ""
+	if len(t.Credentials) > 0 {
```
ehotinger: probably just simplify with `return len (t.Credentials >0)`

graph/task.go
```diff
@@ -197,16 +194,22 @@ func (t *Task) initialize() error {
...
+	}
+	return false
 }
 
 // SetRegistryInfo sets registry information.
 func (t *Task) setRegistryInfo(registry, user, pw string) {
-	t.RegistryName = registry
-	t.RegistryUsername = user
-	t.RegistryPassword = pw
+	t.Credentials = append(t.Credentials, &Credential{
```
ehotinger: where's the struct defined? Make a `NewCredential` function which does validation of all the parameters to ensure they aren't empty; throw an error if they are. Call that method instead of `&Credential`

graph/task_test.go
```diff
@@ -21,10 +21,16 @@ func TestUsingRegistryCreds(t *testing.T) {
...
+		var credentials []*Credential
+
+		credentials = append(credentials, &Credential{
 			RegistryName:     test.registry,
 			RegistryUsername: test.user,
 			RegistryPassword: test.pw,
+		})
+		task := &Task{
+			RegistryName: test.registry,
+			Credentials:  credentials,
```
ehotinger: Simplified as `Credentials: []*Credential{ &Credential { ... } }` - no need for appends or variable. Same for below

graph/task.go
shahzzzam: ah, seems like this request did not take in the file I created for that. Doing it.

builder/builder.go
shahzzzam: Synced with @ehotinger , I will add an enhancement to make it parallel before closing this PR. For now, it won't affect perf a lot because there won't be many credentials

cmd/build.go
```diff
@@ -193,7 +195,24 @@ func (b *buildCmd) createBuildTask() (*graph.Task, error) {
 	// TODO: create secrets
 	secrets := []*graph.Secret{}
 
-	return graph.NewTask(steps, secrets, b.opts.Registry, b.registryUser, b.registryPw, taskTotalTimeoutInSec, true)
+	var credentials []*graph.Credential
+	// If the user provides the username and password, add it to the Credentials
+	if b.opts.Registry != "" && b.registryUser != "" && b.registryPw != "" {
+		creds, _ := graph.NewCredential(b.opts.Registry, b.registryUser, b.registryPw)
```
ehotinger: Even though this can't ever error, just handle it anyways since validation of `NewCredential` could change in the future.

graph/credential.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// Credential defines a combination of registry, username and password
+type Credential struct {
+	RegistryName     string
+	RegistryUsername string
+	RegistryPassword string
+}
+
+// NewCredential creates a new Credential
```
ehotinger: nit: end with period

graph/credential.go
```diff
@@ -0,0 +1,55 @@
...
+package graph
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	errInvalidRegName    = errors.New("registry name can't be empty")
```
ehotinger: nit: caps first letter for consistency?

graph/credential.go
```diff
@@ -0,0 +1,55 @@
...
+// CreateCredentialFromString creates a Credential object from a specified string.
+func CreateCredentialFromString(str string) (*Credential, error) {
+	strs := strings.SplitN(str, ";", 3)
+
+	if len(strs) != 3 {
+		return nil, errInsufficientCreds
+	}
+
+	return NewCredential(strs[0], strs[1], strs[2])
+
```
ehotinger: nit: delete newline

graph/credential_test.go
```diff
@@ -0,0 +1,75 @@
...
+		{"foo;bar;qux;thing", "foo", "bar", "qux;thing", false},
+		{"foo;bar;qux;th;i@ng", "foo", "bar", "qux;th;i@ng", false},
+		{"foo;bar", "foo", "bar", "", true},
+	}
+
+	for _, test := range tests {
+		actual, err := CreateCredentialFromString(test.str)
+		if test.shouldError {
+			if err == nil {
+				t.Fatalf("Test should have errored out,but did not: %v", test)
```
ehotinger: nit: space after `out`

PR #378, opened 2019-01-09
Global Readiness Fixes

djyou:
**Purpose of the PR:**

**Fixes #**


northtyphoon:
The build failed as the test show `country` was still used. Not sure why.

PR #379, opened 2019-01-09
deprecate username/password

shahzzzam:
**Purpose of the PR:**
When `push` flag is specified, the validation checks for source registry, username and password flags to be provided. But, since the `username` and `password` fields are stripped off by runner, and instead added to the `credentials`, we need to update the validation.

ALSO deprecates username/password

**Fixes #**
Update validation to check credential array instead of separate fields.

### DIFF
--- a/cmd/acb/build.go
+++ b/cmd/acb/build.go
@@ -79,7 +79,7 @@ func newBuildCmd(out io.Writer) *cobra.Command {
 	f.StringArrayVar(&r.buildArgs, "build-arg", []string{}, "set build time arguments")
 	f.StringArrayVar(&r.secretBuildArgs, "secret-build-arg", []string{}, "set secret build arguments")
 	f.StringArrayVar(&r.labels, "label", []string{}, "set metadata for an image")
-	f.StringArrayVar(&r.credentials, "credentials", []string{}, "credentials passed on for Source registry plus any custom registries")
+	f.StringArrayVar(&r.credentials, "credentials", []string{}, "credentials passed on for source registry plus any custom registries")
 	f.StringVar(&r.isolation, "isolation", "", "the isolation to use")
 	f.StringVar(&r.target, "target", "", "specify a stage to build")
 	f.StringVar(&r.platform, "platform", "", "sets the platform if the server is capable of multiple platforms")
--- a/cmd/acb/build.go
+++ b/cmd/acb/build.go
@@ -82,8 +82,7 @@ func newBuildCmd(out io.Writer) *cobra.Command {
 	f.StringArrayVar(&r.secretBuildArgs, "secret-build-arg", []string{}, "set secret build arguments")
 	f.StringArrayVar(&r.labels, "label", []string{}, "set metadata for an image")
 
-	// Runner never passed `username` and `password` now. Instead it adds it to `credentials` IFF passed.
-	// It should be safe to remove this from everywhere. But right now, it will just be always empty
+	// `username` and `password` are deprecated. Use --credentials="registryName;username;password" instead
 	f.StringVarP(&r.registryUser, "username", "u", "", "the username to use when logging into the registry")
 	f.StringVarP(&r.registryPw, "password", "p", "", "the password to use when logging into the registry")
 
--- a/cmd/acb/build.go
+++ b/cmd/acb/build.go
@@ -36,8 +36,6 @@ type buildCmd struct {
 	context         string
 	dockerfile      string
 	target          string
-	registryUser    string
-	registryPw      string
 	credentials     []string
 	isolation       string
 	platform        string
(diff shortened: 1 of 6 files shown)


cmd/acb/build.go
```diff
@@ -82,9 +82,12 @@ func newBuildCmd(out io.Writer) *cobra.Command {
 	f.StringArrayVar(&r.secretBuildArgs, "secret-build-arg", []string{}, "set secret build arguments")
 	f.StringArrayVar(&r.labels, "label", []string{}, "set metadata for an image")
 
+	// Runner never passed `username` and `password` now. Instead it adds it to `credentials` IFF passed.
```
northtyphoon: acb doesn't depend on runner. Maybe the comment can just mention the two parameters are deprecated.

cmd/acb/validation.go
```diff
@@ -22,16 +24,22 @@ func validateIsolation(isolation string) error {
...
 
-func validatePush(push bool, registry string, username string, password string) error {
-	if push && (username == "" || password == "" || registry == "") {
-		return errors.New("when specifying push, username, password, and registry are required")
+func validatePush(push bool, credentials []string) error {
+	if push {
+		if len(credentials) == 0 {
+			return errors.New("when specifying push, username, password, and registry are required")
+		}
+		if _, err := graph.CreateCredentialFromString(credentials[0]); err != nil {
```
northtyphoon: Before `validatePush`  is called, has the credential strings already been parsed and validated?

shahzzzam: Done

cmd/acb/validation_test.go
```diff
@@ -49,17 +49,17 @@ func TestValidateRegistryCreds_Invalid(t *testing.T) {
 }
 
 func TestValidatePush_Valid(t *testing.T) {
-	if err := validatePush(false, "", "bar", "qux"); err != nil {
+	if err := validatePush(false, []string{";bar;qux"}); err != nil {
```
northtyphoon: `bar` `foo` seems violating some global readiness policy. Maybe you can change to other strings.

cmd/acb/validation.go
shahzzzam: No, `validatePush` gets called first to check the arguments. All credentials strings are created  when the Build Task is created.

shahzzzam: I think the steps are something like:
1) ValidateCmdArgs (checks for isolation, and push/credentials validation)
2) Create BuildTask (parses everything and creates a task- it parses the strings and passes the creds to Task)

Should I parse the creds as the first step now? and then follow up with the above two steps?

cmd/acb/build.go
```diff
@@ -82,9 +82,11 @@ func newBuildCmd(out io.Writer) *cobra.Command {
 	f.StringArrayVar(&r.secretBuildArgs, "secret-build-arg", []string{}, "set secret build arguments")
 	f.StringArrayVar(&r.labels, "label", []string{}, "set metadata for an image")
 
+	// `username` and `password` are deprecated. Use --credentials="registryName;username;password" instead
 	f.StringVarP(&r.registryUser, "username", "u", "", "the username to use when logging into the registry")
```
ehotinger: just go ahead and delete them so we don't hit similar bugs in the future; it's dead code

cmd/acb/validation.go
```diff
@@ -22,16 +24,22 @@ func validateIsolation(isolation string) error {
 	return nil
 }
 
+// TODO Need to remove this but right now, `username` and `password` are always empty.
```
ehotinger: No TODOs in new code, delete the unused code

cmd/acb/validation.go
ehotinger: just do validation once upon the creation of a Task and ensure that the error is returned to the CLI

cmd/acb/exec.go
```diff
@@ -158,7 +146,7 @@ func (e *execCmd) run(cmd *cobra.Command, args []string) error {
 }
 
 func (e *execCmd) validateCmdArgs() error {
```
ehotinger: delete it entirely

cmd/acb/validation_test.go
```diff
@@ -57,9 +37,3 @@ func TestValidatePush_Valid(t *testing.T) {
 		t.Errorf("All creds are provided, but received an error: %v", err)
 	}
 }
-
-func TestValidatePush_Invalid(t *testing.T) {
```
ehotinger: Should keep the invalid test cases: `true, nil`

cmd/acb/validation.go
```diff
@@ -22,15 +22,8 @@ func validateIsolation(isolation string) error {
...
-		return nil
-	}
-	return errors.New("when specifying username and password, provide both or neither")
-}
-
-func validatePush(push bool, registry string, username string, password string) error {
-	if push && (username == "" || password == "" || registry == "") {
+func validatePush(push bool, credentials []string) error {
+	if push && len(credentials) == 0 {
 		return errors.New("when specifying push, username, password, and registry are required")
```
ehotinger: maybe modify the error: `at least one credential is required`

cmd/acb/build_test.go
```diff
@@ -12,14 +12,13 @@ import (
...
-		dockerfile:   "HelloWorld/Dockerfile",
-		registryUser: "user",
-		registryPw:   "pw",
-		context:      "src",
-		tags:         []string{"foo:latest", "bar/qux"},
-		pull:         true,
-		noCache:      false,
-		dryRun:       true,
+		dockerfile:  "HelloWorld/Dockerfile",
+		credentials: []string{"foo.azurecr.io;user;pw"},
```
ehotinger: you can add a test here to ensure that the buildCmd's specified creds == the generated task's creds

cmd/acb/build.go
```diff
@@ -81,11 +79,7 @@ func newBuildCmd(out io.Writer) *cobra.Command {
 	f.StringArrayVar(&r.buildArgs, "build-arg", []string{}, "set build time arguments")
 	f.StringArrayVar(&r.secretBuildArgs, "secret-build-arg", []string{}, "set secret build arguments")
 	f.StringArrayVar(&r.labels, "label", []string{}, "set metadata for an image")
-
-	f.StringVarP(&r.registryUser, "username", "u", "", "the username to use when logging into the registry")
-	f.StringVarP(&r.registryPw, "password", "p", "", "the password to use when logging into the registry")
-	f.StringArrayVar(&r.credentials, "credentials", []string{}, "all credentials for private repos")
-
+	f.StringArrayVar(&r.credentials, "credentials", []string{}, "credentials passed on for Source registry plus any custom registries")
```
ehotinger: nit: no caps for "Source"

cmd/acb/build_test.go
shahzzzam: we do have that in `credential_test.go`. Is that the one you're talking about?

shahzzzam: done

PR #387, opened 2019-02-08
Add vault packages with functions to retrieve secrets from Azure keyvault using MSI

mnltejaswini:
This PR introduces vaults packages to the repository. This packages will be used to resolve secrets from different types of vaults. Currently added support for getting secrets from Azure Key vaults using the Managed service identity. 

It also add a command to acb called getsecret which can be used to resolve a secret from the supported vaults. However if the secret is in Azure keyvault, it queries the MSI token service running on the host to provide required AAD tokens to access the keyvault.  

Also as a side effect, several vendor packages are updated. 

Tests
------
 Unit tests are added for required functions of azure key vault.

All unit tests passed.
getsecret was able to fetch secret from keyvault on a Azure VM with identity enabled.


cmd/acb/commands/getsecret/getsecret.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"errors"
+	"log"
+
+	"github.com/Azure/acr-builder/vaults"
+	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/urfave/cli"
+)
+
+// Command renders templates and verifies their output.
```
ehotinger: nit: update comment

cmd/acb/commands/getsecret/getsecret.go
```diff
@@ -0,0 +1,64 @@
...
+var Command = cli.Command{
+	Name:  "getsecret",
```
ehotinger: how will you add additional providers to this? maybe consider renaming it or making subgroups?
```
acb decrypt akv --fooFlag
acb decrypt foo --barFlag
```

cmd/acb/commands/getsecret/getsecret.go
```diff
@@ -0,0 +1,64 @@
...
+			Usage: "the azure keyvault secret URL",
+		},
+		cli.StringFlag{
+			Name:  "clientID",
+			Usage: "the MSI user assigned identity client ID",
+		},
+
+		// Rendering options
+		cli.StringFlag{
+			Name:  "vaultAADResourceURL",
```
ehotinger: can we shorten and use all lowercase for all the flag names, maybe

```
akv => --url
clientID => client-id | clientid
vaultADDResourceURL => url | vault | vault-resource-url | ... 
```

vaults/azurekeyvault.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// NewKeyVaultClient creates a new keyvault client
+func newKeyVaultClient(vaultURL, clientID, vaultAADResourceURL string) (*keyVault, error) {
+
+	if vaultAADResourceURL == "" {
+		vaultAADResourceURL = azure.PublicCloud.KeyVaultEndpoint
+	}
+
+	msiKeyConfig := &auth.MSIConfig{
```
northtyphoon: Does the auth library allow overriding the default msi endpoint? /cc: @shahzzzam

vaults/azurekeyvault.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	if secretConfig.VaultURL == "" ||
+		secretConfig.SecretName == "" {
+		return "", errors.New("missing required properties vaultURL and SecretName")
+	}
+
+	keyClient, err := newKeyVaultClient(secretConfig.VaultURL, secretConfig.MSIClientID, secretConfig.AADResourceURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to create azure key vault client. Error: %v", err)
```
ehotinger: you can use `errors.Wrap(err, "failed to create azure key vault client")`

ehotinger: same for below L45

vaults/azurekeyvault.go
```diff
@@ -0,0 +1,139 @@
...
+		vaultURL: vaultURL,
+		client:   &keyClient,
+	}
+
+	return k, nil
+}
+
+// GetSecret retrieves a secret from keyvault
+func (k *keyVault) getSecret(secretName, secretVersion string) (string, error) {
+	ctx := context.Background()
```
ehotinger: you'll want to let the context be passed into this function instead of creating one here, so that timeouts and such can be passed in

cmd/acb/commands/getsecret/getsecret.go
```diff
@@ -0,0 +1,64 @@
...
+		if vaultAADResourceURL == "" {
+			vaultAADResourceURL = azure.PublicCloud.KeyVaultEndpoint
+		}
+
+		secretConfig, err := vaults.NewAKVSecretConfig(akv, clientID, vaultAADResourceURL)
+		if err != nil {
+			return err
+		}
+
+		secretValue, err := secretConfig.GetValue()
```
ehotinger: there's an example of `gocontext` in some of the other commands which just makes a `context.Background()` at this point, would recommend passing that context all the way down

vaults/azurekeyvault.go
mnltejaswini: @northtyphoon  Unfortunately the go SDK for App auth hard codes the MSI endpoint https://github.com/Azure/go-autorest/blob/71d02b67e17a3d80abf8db171f1ca80c39eb2e90/autorest/adal/token.go#L63 

Are we planning to start token service at a different endpoint?

cmd/acb/commands/getsecret/getsecret.go
mnltejaswini: @ehotinger  The word decrypt sounds related to some form of decoding the original secret value. But here the secret is not decoded instead it is stored in a vault and the parameters help to locate them. I am thinking we will have the similar issue with the secrets object in task YAML file. How do we extend to support multiple vault providers? Is it something like 
```
secrets
  -id: akvsecret
    akv: "https://myvault.vault.azure.net/mysecret
  -id: awssecret
    awsv: "<don't know>"
```
So essentially  secrets will have all supported vaults properties and we will resolve them accordingly? If so getsecret can also follow the same pattern, It takes all properties supported by secrets as parameters and resolve them.

vaults/azurekeyvault.go
mnltejaswini: Good point. Will add that.

cmd/acb/commands/getsecret/getsecret.go
mnltejaswini: I will add the context but planning to have command line wait till the underlying SDK calls returns as the autorest has some decent timeouts and retry policies.

docs/getsecret_akv.md
```diff
@@ -0,0 +1,110 @@
+# GetSecrets from Azure Keyvault using MSI
+
+acr-builder has added support to get secrets from azure keyvault using managed service dientity (https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview). It has a new command called **getsecret** that can fetch secret from different vault providers. Currently it supports secrets from Azure key vault. The below article describes how to use acb to retrieve a secret from Azure keyvault using Managed Identities.
```
shahzzzam: NIT: identity*

PR #389, opened 2019-02-12
Adds support of secrets to task template

mnltejaswini:
This PR adds the support of secrets to template. Currently the secrets can be retrieved from Azure keyvault using MSI. The resolved secrets can then be used in the known templating language like {{.Secrets.mysecret}} The documentation for using the secrets is added to the repo. 

Tests
----
  Unit tests cases added to all new and updated modules of the repository. Existing unit tests passed. E2E test on a Azure VM using MSI is documented.


templating/secrets.go
```diff
@@ -0,0 +1,126 @@
...
+		// Block until either:
+		// - The global context expires
+		// - Resolving a secret has error
+		// - All secrets are resolved successfully
+		for _, ch := range resolvedChans {
+			select {
+			case <-ctx.Done():
+				return resolvedSecrets, ctx.Err()
+			case secretValue := <-ch:
+				resolvedSecrets[secretValue.ID] = secretValue.Value
```
northtyphoon: How does the code handle two secrets with the same id? Will it fail somewhere in the pipeline or just overwrite?

templating/secrets.go
```diff
@@ -0,0 +1,126 @@
...
+			case err := <-errorChan:
```
northtyphoon: The error might not come from current secret resolving job since it's shared. But I assume the code just wants to bail out  whenever an error occurs.

templating/secrets.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/pkg/errors"
+)
+
+// ResolveSecretFunc is a function that resolves the secret to its value and sends through the ResolvedChan of the secret. Any errors during resolve are send through errorChan
+type ResolveSecretFunc func(ctx context.Context, azureVaultResourceURL string, secret *graph.Secret, errorChan chan error)
+
+// SecretResolver defines how a secret is resolved.
+type SecretResolver struct {
+	Resolve          ResolveSecretFunc
+	azureEnvironment azure.Environment
```
northtyphoon: SecretResolver sounds like a generic and not bound to azure. Maybe AzureSecretResolver?

templating/base_render_options.go
```diff
@@ -152,6 +159,54 @@ func LoadAndRenderSteps(template *Template, opts *BaseRenderOptions) (string, er
 	return rendered, nil
 }
 
+// ResolveAndMergeSecrets parses the secrets in the template, resolves them using vault providers and adds them to the values map.
+func resolveAndMergeSecrets(ctx context.Context, template *Template, secretResolver *SecretResolver, valueMap *Values) error {
```
northtyphoon: Do you need to pass *Values or maybe just Values? I also think return (Values, error) is more clear.

graph/secret.go
```diff
@@ -1,10 +1,76 @@
...
+}
+
+// Secret defines a wrapper to resolve vault secrets to values.
 type Secret struct {
-	Akv       string `yaml:"akv,omitempty"`
-	SecretEnv string `yaml:"secretEnv,omitempty"`
+	ID          string `yaml:"id"`
+	Akv         string `yaml:"akv,omitempty"`
+	MsiClientID string `yaml:"client-id,omitempty"`
+
```
northtyphoon: I think we avoid using `-` in Task. https://docs.microsoft.com/en-us/azure/container-registry/container-registry-tasks-reference-yaml

cmd/acb/commands/exec/exec.go
```diff
@@ -186,7 +191,13 @@ var Command = cli.Command{
 			}
 		}
 
-		rendered, err := templating.LoadAndRenderSteps(template, renderOpts)
+		secretResolver, err := templating.DefaultSecretResolver(azureEnvironmentName)
+		if err != nil {
+			return err
+		}
+
+		// TODO: Do we need to add any timeout with the ctx for resolving the secrets?
```
northtyphoon: Yes. Better to have a timeout in case it's stuck somewhere. Otherwise users won't see any output from stream logs until the client side timeout.

graph/secret.go
```diff
@@ -1,10 +1,76 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
ehotinger: keep headers

graph/secret.go
```diff
@@ -1,10 +1,76 @@
...
+// Secret defines a wrapper to resolve vault secrets to values.
 type Secret struct {
-	Akv       string `yaml:"akv,omitempty"`
-	SecretEnv string `yaml:"secretEnv,omitempty"`
+	ID          string `yaml:"id"`
+	Akv         string `yaml:"akv,omitempty"`
+	MsiClientID string `yaml:"client-id,omitempty"`
+
+	// ResolvedChan is used to signal the callers
+	// that the secret has been resolved successfully to a vaule.
```
ehotinger: nit: typo

graph/secret.go
```diff
@@ -1,10 +1,76 @@
...
+
+// Equals determines whether or not two secrets are equal.
+func (s *Secret) Equals(t *Secret) bool {
+	if s == nil && t == nil {
+		return true
+	}
+	if s == nil || t == nil {
+		return false
+	}
+	if s.ID != t.ID ||
```
ehotinger: nit: can just put return here instead of if/else

graph/task.go
```diff
@@ -80,6 +80,16 @@ func UnmarshalTaskFromFile(file string, creds []*Credential) (*Task, error) {
 	return t, err
 }
 
+// NewTaskFromString unmarshals a Task from string without any initialization.
+func NewTaskFromString(data string) (*Task, error) {
+	t := &Task{}
```
ehotinger: nit: a bit cleaner if you just return the error instead of wrapping it here; you end up handling it twice right now

cmd/acb/commands/render/render.go
```diff
@@ -129,7 +135,12 @@ var Command = cli.Command{
 			}
 		}
 
-		rendered, err := templating.LoadAndRenderSteps(template, renderOpts)
+		secretResolver, err := templating.DefaultSecretResolver(azureEnvironmentName)
```
ehotinger: no reason to do the allocation here when it may not be used at all

ehotinger: same for exec

templating/base_render_options.go
```diff
@@ -152,6 +159,54 @@ func LoadAndRenderSteps(template *Template, opts *BaseRenderOptions) (string, er
...
+
+	// Cheap optimization to skip the secrets merging if it doesn't contain "secrets" string in it. Note that the task can
+	// have the string secrets but may not essentially the secrets section.
+	if !strings.Contains(string(template.Data), "secrets") {
```
ehotinger: hmm... if we are optimizing for no secrets perhaps we should do the deserialization first and simply check len(deserialized secrets) > 0 so that we don't have to do a full scan here. Of course that will result in more allocations than in the has secrets path. Can we get a better hot/cold path from the CLI parameters?

templating/base_render_options.go
```diff
@@ -152,6 +159,54 @@ func LoadAndRenderSteps(template *Template, opts *BaseRenderOptions) (string, er
...
+		return nil
+	}
+
+	// At first render the template with existing values to render templatized vaules for secrets.
```
ehotinger: nit: typo

templating/base_render_options.go
```diff
@@ -152,6 +159,54 @@ func LoadAndRenderSteps(template *Template, opts *BaseRenderOptions) (string, er
...
+	values := *valueMap
```
ehotinger: I guess copying the pointer doesn't do much here

templating/base_render_options.go
```diff
@@ -152,6 +159,54 @@ func LoadAndRenderSteps(template *Template, opts *BaseRenderOptions) (string, er
...
+	values["Secrets"] = Values{}
+	engine := NewEngine()
```
ehotinger: nit: can save an allocation on the engine by re-using it between the double render

templating/base_render_options.go
```diff
@@ -152,6 +159,54 @@ func LoadAndRenderSteps(template *Template, opts *BaseRenderOptions) (string, er
...
+	// If no secrets found return.
+	if len(task.Secrets) == 0 {
+		return nil
+	}
+
+	if secretResolver == nil {
+		return errors.New("secret resolver is required")
+	}
+
+	resolvedSecrets, err := secretResolver.ResolveSecrets(ctx, task.Secrets)
```
ehotinger: seems like the instantiation of the secret resolver should go before this line so that it's not unnecessarily allocated

templating/secrets.go
```diff
@@ -0,0 +1,126 @@
...
+
+// DefaultSecretResolver creates a default secret resolver for the given azure environment
+func DefaultSecretResolver(azureEnvironmentName string) (*SecretResolver, error) {
+	return NewSecretResolver(resolveSecret, azureEnvironmentName)
+}
+
+// ResolveSecrets returns a list of resolved secrets and returns error if there is any failure in resolving a secret.
+func (secretResolver *SecretResolver) ResolveSecrets(ctx context.Context, secrets []*graph.Secret) (Values, error) {
+
+	resolvedSecrets := Values{}
```
ehotinger: nit: avoid an allocation by placing under the guard. Also, the guard doesn't really guard anything since it's guaranteed to be non-zero from the code path right now, but I don't mind defensive functions 👍

templating/secrets.go
ehotinger: yea, race for the overwrite - maybe you can add validation during task creation for it

templating/secrets.go
```diff
@@ -0,0 +1,126 @@
...
+
+		if endIndex > len(secrets) {
+			endIndex = len(secrets)
+		}
+
+		var resolvedChans []chan graph.SecretValue
+
+		for _, secret := range secrets[index:endIndex] {
+			// Validate secret before resolving
+			err := secret.Validate()
```
ehotinger: move all the validation to during task creation so that it's unified in one place?

templating/secrets.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+		}
+	}
+
+	return resolvedSecrets, nil
+}
+
+func resolveSecret(ctx context.Context, azureVaultResourceURL string, secret *graph.Secret, errorChan chan error) {
+	if secret == nil {
+		errorChan <- errors.New("secret cannot be nil")
```
ehotinger: what's the point of the channel, why not just return an error?

templating/secrets.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	if secret.IsAkvSecret() {
+		secretConfig, err := vaults.NewAKVSecretConfig(secret.Akv, secret.MsiClientID, azureVaultResourceURL)
+		if err != nil {
+			errorChan <- errors.Wrap(err, "failed to create azure keyvault secret config")
```
ehotinger: you would want to `return` after all of these channel writes, I would recommend not using channels here though

ehotinger: bug: return after all of these channel writes, the code will continue execution right now. same for below

templating/secrets.go
mnltejaswini: My understanding was the last write will be success just like we override the values? If not I can fail the task if it has duplicate IDs

templating/secrets.go
mnltejaswini: SecretResolver is generic and should be able abstract from the vault it is being resolved

graph/secret.go
mnltejaswini: clientID or msiClientID?

templating/secrets.go
mnltejaswini: Thanks a lot for catching, will add a unit test for it

templating/base_render_options.go
mnltejaswini: If we instantiate the secretresolver over here, there are two concerns. 1. We need to pass all initialization parameters for secret resolver to this function 2. If we hard code the secret resolver, it will be hard to unit test this function by mocking the resolver

templating/secrets.go
mnltejaswini: Yes, if we cannot resolve any secrets we will fail the task

templating/secrets.go
ehotinger: personally I'm ok with leaving it as-is, maybe make an issue to track it

templating/base_render_options.go
mnltejaswini: Not sure I understood your comment. Can you please elaborate?

graph/secret.go
ehotinger: `clientID` seems fine, we can leave the property as-is and refactor it if we support multiple in the future

templating/secrets.go
northtyphoon: I see. If SecretResolver is designed to be generic interface for all providers, it probably should not include azureEnvironment. Ideally ResolveSecretFunc can abstract it. Or you can introduce another AzureSecretResolver based on SecretResolver. 
It's not critical. I think we can address it later.

templating/secrets.go
northtyphoon: Override is fine as long as the result is predictable and consistent. Since the secret resolver is not executed in order, the result may not be consistent.

graph/task.go
```diff
@@ -62,24 +62,79 @@ func UnmarshalTaskFromString(data string, defaultWorkDir string, network string,
...
+		if _, exists := idMap[secret.ID]; exists {
+			return fmt.Errorf("duplicate secret found with ID: %s", secret.ID)
+		}
+
+		idMap[secret.ID] = struct{}{}
+	}
+	// Validate steps if exists
+	idMap = make(map[string]struct{}, len(task.Steps))
+	for _, step := range task.Steps {
+		err := step.Validate()
```
mnltejaswini: @ehotinger  I guess this will regress the current behavior right?

ehotinger: yeah, just do validation after initialization if you want to keep this function, you could also just add all of this to the initialize func

mnltejaswini: We talked offline and for now we will keep only secret validation as part of task validation

PR #402, opened 2019-03-06
allow overriding the default msi endpoint

northtyphoon:
**Purpose of the PR:**
Windows container handles 169.254.169.254 as local-link address and only route to on-link device by default. As the limitation, to allow acb talks to the msi token server on a different address, the change introduces a environment variable to override the default endpoint.

### DIFF
--- a/vaults/azurekeyvault.go
+++ b/vaults/azurekeyvault.go
@@ -19,7 +19,7 @@ import (
 
 const (
 	// environment variable to override the default msi endpoint
-	envAcrMsiEndpoint = "ACR_MSI_ENDPOINT"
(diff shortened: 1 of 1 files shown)


vaults/azurekeyvault.go
```diff
@@ -140,3 +148,31 @@ func (k *keyVault) getSecret(ctx context.Context, secretName, secretVersion stri
 
 	return *secretBundle.Value, nil
 }
+
+// newAuthorizer creates the authorizer for kev vault client.
```
shahzzzam: nit: keyvault*

PR #409, opened 2019-03-18
Start validating version

### LINKED ISSUE #397: Validate & log version
Validate the version is <= default version and log the version when a task is being executed.

ehotinger:
**Purpose of the PR:**
Start validating/enforcing the version attribute on a Task.

There are currently two valid values:
- v1.0.0 (default)
- [1.0-preview-1](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-tasks-multi-step)

**Fixes #397**

### DIFF
--- a/graph/task.go
+++ b/graph/task.go
@@ -24,6 +24,16 @@ const (
 
 	// The default step retry delay is 5 seconds.
 	defaultStepRetryDelayInSeconds = 5
+
+	// currentTaskVersion is the most recent Task version.
+	currentTaskVersion = "v1.0.0"
+)
+
+var (
+	validTaskVersions = map[string]bool{
+		"1.0-preview-1":    true,
+		currentTaskVersion: true,
+	}
 )
 
 // Task represents a task execution.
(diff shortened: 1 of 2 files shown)


graph/task.go
```diff
@@ -24,6 +24,16 @@ const (
 
 	// The default step retry delay is 5 seconds.
 	defaultStepRetryDelayInSeconds = 5
+
+	// currentTaskVersion is the most recent Task version.
+	currentTaskVersion = "v1.0.0"
```
northtyphoon: Does it require to add `v` prefix for semantic version?

ehotinger: Nah, from FAQ of SemVer:

> Is "v1.2.3" a semantic version?
> No, "v1.2.3" is not a semantic version. However, prefixing a semantic version with a "v" is a common way (in English) to indicate it is a version number. Abbreviating "version" as "v" is often seen with version control. Example: git tag v1.2.3 -m "Release version 1.2.3", in which case "v1.2.3" is a tag name and the semantic version is "1.2.3".
> 

I think it makes sense to have the v prefix here though, since it's consistent with other projects and we can re-use it for the git tag (+automatic linking during compile time via the existing `version` tag), and the vx.x.x is the pattern for vgo/modules.

ehotinger: https://github.com/golang/go/wiki/Modules#version-selection

PR #410, opened 2019-03-18
Add YAML docs for Task

### LINKED ISSUE #396: Docs: Add/Update schema docs for yaml + templates
Add/Update schema docs for yaml + templates

ehotinger:
**Purpose of the PR:**

- Add YAML docs for Task

**Fixes #396**

### DIFF
--- a/docs/task.md
+++ b/docs/task.md
@@ -1,0 +1,391 @@
+# Task Schema
+
+## Task Properties
+
+| Property | Type | Required | Nullable |
+|----------|------|----------|----------|
+| [steps](#steps) | `step[]` | Required  | No |
+| [stepTimeout](#steptimeout) | `int` | Optional  | Yes |
+| [totalTimeout](#totaltimeout) | `int` | Optional | Yes |
+| [secrets](#secrets) | `secret[]` | Optional | Yes |
+| [networks](#networks) | `network[]` | Optional | Yes |
+| [envs](#envs) | `string[]` | Optional | Yes |
+| [workingDirectory](#workingdirectory) | `string` | Optional | Yes |
+| [version](#version) | `string` | Optional | Yes |
+
+## steps
+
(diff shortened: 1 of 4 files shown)


ehotinger:
Open to feedback on this, expect it to be relatively iterative. /cc @sajayantony 

@mnltejaswini  + @shahzzzam  can you also suggest improvements to docs for secrets. I know that area is mostly still a WIP so it can be a rough skeleton for now.

shahzzzam:
finally :D

docs/task.md
```diff
@@ -0,0 +1,391 @@
...
+#### startDelay
+
+Defines the number of seconds to wait before executing the container.
+
+* Optional
+* Type: `int`
+
+#### retries
+
+The number of retries to attempt if a container fails its execution.
```
northtyphoon: Can you elaborate how we determine containers fails? eg, return code is not 0

docs/task.md
```diff
@@ -0,0 +1,391 @@
...
+* Optional
+* Type: `string`
+
+### step
+
+An object with the following properties:
+
+| Property | Type | Required | Nullable |
+|----------|------|----------|----------|
+| [id](#id) | `string` | Optional | Yes |
```
northtyphoon: What does Nullable mean? Especially for int or bool type. I think we probably need a default value column.

docs/task.md
```diff
@@ -0,0 +1,391 @@
+# Task Schema
```
northtyphoon: add a version in the header

docs/task.md
ehotinger: Removed, added default value

docs/task.md
ehotinger: Done

docs/task.md
ehotinger: Done

PR #412, opened 2019-03-18
Breaking change: rename envs to env

ehotinger:
**Purpose of the PR:**

Breaking change: rename `env` to `envs` to match pluralization on `Step` with `Task`.

### DIFF
--- a/docs/task.md
+++ b/docs/task.md
@@ -4,7 +4,7 @@
 
 `v1.0.0`
 
-## Task Properties
+## Task
 
 | Property | Type | Required | Default Value |
 |----------|------|----------|---------------|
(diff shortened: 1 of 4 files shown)


sajayantony:
I'm going by convention here. Docker compose and kube as well use env. Lets discuss this before going forward with this change.

sajayantony:
The real change i want was passing in a --env-file or could have come from secret args here as an example - 
https://github.com/sajayantony/task-webapp/blob/master/task/acb.yaml#L4

ehotinger:
@sajayantony could you make an issue for the environment file (it sounds similar to Node) and jot down your thoughts on how templating will work with it (if at all)?

SteveLasker:
I like the env file approach as well.

ehotinger:
@SteveLasker maybe you could make the issue / design for it?

SteveLasker:
It's now in our backlog

Issue #414, opened 2019-03-20
Login to custom registries using Identity or vault secrets


shahzzzam:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

**What happened?**:
Add ability to support vault secrets for username and password
Add ability to login to a Azure Container Registry using Azure Managed Identity

**What did you expect to happen?**:

**How do you reproduce it (as minimally and precisely as possible)?**:

**Environment**:
All environments 

**Anything else we need to know?**:

Extension to #361 which was meant for plainetxt credentials only

PR #415, opened 2019-03-20
Add ability to resolve custom credential secrets of Vault/MSI type

shahzzzam:
**Purpose of the PR:**
This adds the feature of resolving custom credentials of a Registry, which may be in the form of Vault or MSI, and login to the Registry.

**Fixes #**
#414


builder/builder.go
```diff
@@ -60,14 +68,22 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
...
+			return err
+		}
+
+		for _, cred := range resolvedCreds {
 			loginCtx, cancel := context.WithTimeout(ctx, timeout)
 			defer cancel()
-			log.Printf("Logging in to registry: %s\n", cred.RegistryName)
-			if err := b.dockerLoginWithRetries(loginCtx, cred.RegistryName, cred.RegistryUsername, cred.RegistryPassword, 0); err != nil {
+			log.Printf("Logging in to registry: %s\n", cred.name)
+			// add logic for grabbing all credentials.
```
northtyphoon: is the comment still relevant?

builder/builder.go
```diff
@@ -379,3 +395,82 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+			resolvedCreds[usernameKey] = "00000000-0000-0000-0000-000000000000"
+			unresolvedCreds = append(unresolvedCreds, &graph.Secret{
+				InternalID:    passwordKey,
+				MsiClientID:   cred.Identity,
+				ArmResourceID: cred.ArmResourceID,
+			})
+		}
+	}
+
+	secretResolver, err := templating.NewSecretResolver(nil, templating.DefaultSecretResolveTimeout)
```
northtyphoon: You may consider to refactor SecretResolver to a separate module. It's strange to reference anything from templating in this function.

graph/registry_credential.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if registry == "" {
+		return nil, errInvalidRegName
+	}
+
+	var retVal *RegistryCredential
+
+	isOpaque := usernameType == Opaque && passwordType == Opaque
+	hasVaultSecret := usernameType == VaultSecret || passwordType == VaultSecret
+	isMSI := usernameType == "" || passwordType == ""
```
northtyphoon: both need to be empty?

graph/registry_credential.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/pkg/errors"
+)
+
+var (
+	errInvalidRegName       = errors.New("registry name can't be empty")
+	errInvalidUsername      = errors.New("username can't be empty")
+	errInvalidPassword      = errors.New("password can't be empty")
+	errInvalidIdentity      = errors.New("identity can't be empty")
+	errInvalidArmResourceID = errors.New("armResourceId can't be empty")
+	errSomethingWrong       = errors.New("something went wrong")
```
northtyphoon: the error message won't help much

builder/builder.go
```diff
@@ -379,3 +395,82 @@ func parseImageNameFromArgs(cmdArgs string) string {
 	}
 	return cmdArgs[:idx]
 }
+
+func resolveCustomRegistryCredentials(ctx context.Context, credentials []*graph.RegistryCredential) ([]*ResolvedRegistryCred, error) {
```
northtyphoon: The code seems too complicated. I'd expect the main logic is like 
```
for _, cred := range credentials {
   if msi, get username, password from msi
   if keyvault, get username, password from keyvault
   if opaque, get username, password from the raw string
}
```

graph/secret.go
```diff
@@ -27,6 +27,14 @@ type Secret struct {
 	Akv         string `yaml:"akv,omitempty"`
 	MsiClientID string `yaml:"clientID,omitempty"`
 
+	// InternalID is an ID generated by secrets not specified in Task yaml file.
+	// For instance, Secrets passed in for login to custom registry will generate internal IDs.
+	// Need a separate ID so that it does not conflict with user-defined IDs
+	InternalID string
```
northtyphoon: secret is an entity in task. it's very different from the registry credentials even though both can reference key vault. so I don't think it's a right way to extend this secret model to support registry credential.

tokenutil/token_util.go
```diff
@@ -0,0 +1,117 @@
...
+type RegistryRefreshToken struct {
+	RefreshToken string `json:"refresh_token"`
+}
+
+// GetRegistryRefreshToken gets ARM token, and makes a POST request to registry/v2 endpoint to
+// exchange it for Registry token. Steps mentioned in detail below
+// Authentication: https://github.com/Azure/acr/blob/master/docs/AAD-OAuth.md#authenticating-to-a-registry-with-azure-cli
+// Exchange: https://github.com/Azure/acr/blob/master/docs/AAD-OAuth.md#calling-post-oauth2exchange-to-get-an-acr-refresh-token
+// Note, we don't need to do token challenge part.
+func GetRegistryRefreshToken(registry, resourceID, clientID string) (string, error) {
```
northtyphoon: AccessToken

secretmgmt/secret.go
```diff
@@ -1,35 +1,35 @@
...
 type Secret struct {
 	ID          string `yaml:"id"`
 	Akv         string `yaml:"akv,omitempty"`
 	MsiClientID string `yaml:"clientID,omitempty"`
 
+	// After the Secret is resolved, the value can be found here.
+	ResolvedValue string
+
+	// ArmResourceID is used to fetch ARM token from a TokenServer for an identity
+	ArmResourceID string
```
shahzzzam: Should we add this ArmResourceID to include MSI secrets here? Just how Akv exists as a part of "Secret"
Or should we split this into two classes: `MSISecret` and `VaultSecret`.
Since when it comes to MSI secrets `ARMResourceID`, we have full control over the `ID` property here. Since it is not entered by user. Right now, I use it to send `RegistryName` so we can obtain a token for the registry. What I could do is make the `ID` "registryName_armResourceId` and delete this field. 
@northtyphoon  @ehotinger

tokenutil/token_util.go
shahzzzam: @northtyphoon as I mentioned yesterday. ACR gives back a `refresh_token`. This is similar code as in the CLI

graph/registry_credential.go
```diff
@@ -0,0 +1,115 @@
...
+	PasswordType  string `json:"passwordProviderType,omitempty"`
+	Identity      string `json:"identity,omitempty"`
+	ArmResourceID string `json:"armResourceId,omitempty"`
+}
+
+// CreateRegistryCredentialFromString creates a RegistryCredential object from a serialized string.
+func CreateRegistryCredentialFromString(str string) (*RegistryCredential, error) {
+	var cred RegistryCredential
+	if err := json.Unmarshal([]byte(str), &cred); err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("unable to unmarshal Credentials from string '%s'", str))
```
ehotinger: can't return an error with the string here since it might contain sensitive information -- also there's Wrapf

graph/registry_credential.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	registry := cred.Registry
+	username := cred.Username
+	usernameType := strings.ToLower(cred.UsernameType)
+	password := cred.Password
+	passwordType := strings.ToLower(cred.PasswordType)
+	identity := cred.Identity
+	armResourceID := cred.ArmResourceID
```
ehotinger: a lot of these are pointless variables, just use cred directly

graph/registry_credential_test.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package graph
+
+import "testing"
+
+func TestCreateCredentialFromString(t *testing.T) {
```
ehotinger: can you add tests for the MSI and default case as well, get 100% coverage

secretmgmt/secrets.go
```diff
@@ -19,12 +19,12 @@ const (
 )
```
ehotinger: nit: move to a consts file - `consts` or `appdefaults` would be nice

cmd/acb/commands/build/build_test.go
```diff
@@ -60,10 +60,10 @@ func TestCreateBuildTask(t *testing.T) {
 	}
 
 	taskCreds := task.Credentials[0]
-	expectedCreds, _ := graph.CreateCredentialFromString(creds[0])
-	if taskCreds.RegistryName != expectedCreds.RegistryName ||
-		taskCreds.RegistryUsername != expectedCreds.RegistryUsername ||
-		taskCreds.RegistryPassword != expectedCreds.RegistryPassword {
+	expectedCreds, _ := graph.CreateRegistryCredentialFromString(creds[0])
```
ehotinger: handle error

cmd/acb/commands/build/build_test.go
```diff
@@ -60,10 +60,10 @@ func TestCreateBuildTask(t *testing.T) {
...
+	if taskCreds.Registry != expectedCreds.Registry ||
```
ehotinger: add an Equals method to the credential class and invoke it here

secretmgmt/secrets.go
```diff
@@ -41,15 +41,11 @@ func NewSecretResolver(resolveFunc ResolveSecretFunc, resolveTimeout time.Durati
 	return &SecretResolver{Resolve: resolveFunc, resolveTimeout: resolveTimeout}, nil
 }
 
-// ResolveSecrets returns a list of resolved secrets and returns error if there is any failure in resolving a secret.
-func (secretResolver *SecretResolver) ResolveSecrets(ctx context.Context, secrets []*graph.Secret) (Values, error) {
-
-	resolvedSecrets := Values{}
-
+// ResolveSecrets returns a map of resolved secrets (secretID:secretValue), and returns error if there is any failure in resolving a secret.
```
ehotinger: could you explain why you wanted to shift to using a bool instead of writing the value and not return the list of values and instead set it higher up?
also this comment is out of date

shahzzzam: updated the comment.
As for using bool, we just wanted a signal to say the goroutine has been completed. We set the resolved value in the goroutine itself...

PR #425, opened 2019-03-23
add docs, make couple bugfixes

shahzzzam:
**Purpose of the PR:**
- Adds more documentation for custom credentials.
- Rename `armResourceId` to `aadResourceId`.
- Remove `az-cloud-name`

**Fixes #**
- Fix a bug where it couldn't make POST request using `client.Do`. So using `client.Post` instead.

### DIFF
--- a/docs/custom_registry_login.md
+++ b/docs/custom_registry_login.md
@@ -149,7 +149,7 @@ The idea is:
 root@testacbvm:~# docker run -v $(pwd):/workspace --workdir /workspace -v /var/run/docker.sock:/var/run/docker.sock acb exec  \
 --homevol $(pwd) -f mytask.yaml \
 --credential '{"registry":"myregistry1.azurecr.io","userNameProviderType":"vaultsecret","username":"https://myacbvault.vault.azure.net/secrets/username","passwordProviderType":"vaultsecret","password":"https://myacbvault.vault.azure.net/secrets/password","identity":"c72b2df0-b9d8-4ac6-9363-7c1eb06c1c86"}' \
---credential '{"registry":"myregistry2.azurecr.io","identity":"c72b2df0-b9d8-4ac6-9363-7c1eb06c1c86","armResource":"https://management.azure.com/"}'
+--credential '{"registry":"myregistry2.azurecr.io","identity":"c72b2df0-b9d8-4ac6-9363-7c1eb06c1c86","aadResourceId":"https://management.azure.com/"}'
 
 
 2019/03/23 00:58:07 Using /home/azureuser as the home volume
--- a/docs/custom_registry_login.md
+++ b/docs/custom_registry_login.md
@@ -1,0 +1,217 @@
+# Custom registry login using Identity
+
+acr-builder has added support to login to multiple custom registries using managed service identity (https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview). The below article describes how to use acb to retrieve a login credentials from Azure keyvault using Managed Identities or directly logging in to the registry using Managed Identity. 
+
+We will use **exec** command of the acr-builder to execute acr task with specified values
+
+```
+NAME:
+   acb.exe exec - execute a task file
+
+USAGE:
+   acb.exe exec [command options] [arguments...]
+
+OPTIONS:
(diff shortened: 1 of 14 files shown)


docs/custom_registry_login.md
```diff
@@ -0,0 +1,218 @@
+# Custom registry login using Identity
```
sajayantony: Generic feedback. Start sentences with Upper case.

docs/custom_registry_login.md
```diff
@@ -0,0 +1,218 @@
...
+### Create Azure VM with identity
+
+```
+az vm create -n testacbvm -g myResourceGroup --image UbuntuLTS --assign-identity $ID
+```
+
+### SSH to VM and set up with docker and acb image
+
+```
+sudo -s
```
sajayantony: As a best practice try not dropping to root and just run `sudo apt-get install docker-ce -y`

graph/registry_credential.go
```diff
@@ -15,7 +15,7 @@ var (
 	errInvalidUsername      = errors.New("username can't be empty")
 	errInvalidPassword      = errors.New("password can't be empty")
 	errInvalidIdentity      = errors.New("identity can't be empty")
-	errInvalidArmResourceID = errors.New("armResourceId can't be empty")
```
sajayantony: Recommend renaming this to `resourceId` ref  https://github.com/Azure/azure-sdk-for-go/blob/1951233eb944a49aa5f7278cba8e3e32a8c958af/services/resources/mgmt/2018-05-01/resources/resources.go#L137

graph/registry_credential_test.go
```diff
@@ -33,7 +33,7 @@ func TestCreateCredentialFromString(t *testing.T) {
 		{`{"registry":"r","identity":"clientID", "armResourceId": "https://management.azure.com"}`, true, &RegistryCredential{
 			Registry:      "r",
 			Identity:      "clientID",
-			ArmResourceID: "https://management.azure.com",
+			ArmResource: "https://management.azure.com",
```
sajayantony: Could we just use the azure SDK DefaultBaseURI definition - https://github.com/Azure/azure-sdk-for-go/blob/master/services/resources/mgmt/2018-05-01/resources/client.go#L29

northtyphoon: Strictly speaking, baseuri is different from resource registered in AAD. For most resources, they might be same but arm is a special case.
Here, the string is used for test and can be anything. It doesn't really matter.

graph/registry_credential_test.go
```diff
@@ -33,7 +33,7 @@ func TestCreateCredentialFromString(t *testing.T) {
 		{`{"registry":"r","identity":"clientID", "armResourceId": "https://management.azure.com"}`, true, &RegistryCredential{
```
northtyphoon: still `armResourceId`?

graph/registry_credential.go
```diff
@@ -93,13 +93,13 @@ func CreateRegistryCredentialFromString(str string) (*RegistryCredential, error)
...
 		}
-		if cred.ArmResourceID == "" {
+		if cred.ArmResource == "" {
 			return nil, errInvalidArmResourceID
 		}
 		retVal = &RegistryCredential{
 			Registry:      cred.Registry,
 			Identity:      cred.Identity,
-			ArmResourceID: cred.ArmResourceID,
+			ArmResource: cred.ArmResource,
```
northtyphoon: nit: format

graph/registry_credential_test.go
sajayantony: The property at this point is not self describing. Is it’s a URI or a resourceId or an endpoint etc.

docs/custom_registry_login.md
shahzzzam: I wanted to be consistent with some of our other docs:
https://github.com/Azure/acr-builder/blob/38f06e5acb94854c74f6fb03f7958dc711dfbd90/docs/task-with-secrets.md#ssh-to-vm-and-set-up-with-docker-and-acb-image

shahzzzam: but you're right, I updated it.

PR #429, opened 2019-03-28
log commitID and branch information

shahzzzam:
**Purpose of the PR:**
In the download.go, we fetch `CommitID` and `Branch` information of the Git context.
And we log it. 

Reason behind logging it is, it can be fetched by the STDOUT buffers for the client to fetch Git commit,branch information.

**Fixes #**

### DIFF
--- a/cmd/acb/commands/download/download.go
+++ b/cmd/acb/commands/download/download.go
@@ -5,6 +5,7 @@ package download
 
 import (
 	gocontext "context"
+	"encoding/json"
 	"errors"
 	"log"
 	"time"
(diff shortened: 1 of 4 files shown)


scan/context.go
```diff
@@ -29,17 +29,18 @@ const (
...
+func (s *Scanner) ObtainSourceCode(ctx context.Context, scContext string) (workingDir string, sha string, branch string, err error) {
 	isGitURL, workingDir, err := s.getContext(scContext)
 	if err != nil {
-		return workingDir, sha, err
+		return workingDir, sha, branch, err
 	}
 
 	if isGitURL {
 		sha, err = s.GetGitCommitID(ctx, workingDir)
+		branch, err = s.GetGitBranchName(ctx, workingDir)
```
ehotinger: short-circuit here if commit id fails

cmd/acb/commands/download/download.go
```diff
@@ -14,6 +15,11 @@ import (
 	"github.com/urfave/cli"
 )
 
+type gitInfo struct {
+	CommitID string
```
ehotinger: add json properties here so that it's clear what they'll serialize as - i.e. `commitID` vs `commitId`

PR #431, opened 2019-03-30
add support for working-directory for local build context

shahzzzam:
Allows passing in `working-directory` flag to build. This is to support building local context.

### DIFF
--- a/cmd/acb/commands/build/build_test.go
+++ b/cmd/acb/commands/build/build_test.go
@@ -29,9 +29,10 @@ func TestCreateBuildTask(t *testing.T) {
 		renderOpts      = &templating.BaseRenderOptions{
 			Registry: registry,
 		}
-		debug = false
-		push  = true
-		creds = []string{`{"registry":"foo.azurecr.io","username":"user","userNameProviderType":"opaque","password":"pw","passwordProviderType":"opaque"}`}
+		debug      = false
+		push       = true
+		creds      = []string{`{"registry":"foo.azurecr.io","username":"user","userNameProviderType":"opaque","password":"pw","passwordProviderType":"opaque"}`}
+		workingDir = ""
 	)
 
 	task, err := createBuildTask(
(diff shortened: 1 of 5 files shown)


scan/git.go
```diff
@@ -26,6 +26,16 @@ type gitRepo struct {
 	subdir string
 }
 
+// IsGitRepo checks whether current directory is a GIT directory
+func (s *Scanner) IsGitRepo(ctx context.Context, cmdDir string) bool {
```
ehotinger: IMO remove this code path and just pass in the `--branch` and `--commit` to avoid running the commands 2x. If those are specified with a local context you can populate the dependencies from it

ehotinger: I think that'll ensure we're consistent with the branch and commit rendering since they can muck with the context at this point, checking out other branches, etc.

shahzzzam: This is done to also support passing in remote context.

Like `acb build https://github.com/blah.git`

cc @northtyphoon

ehotinger: I know... let me if the comment doesn't make sense and I can rephrase it.

northtyphoon: If we want to completely avoid calling the commands 2x, pass --commit all the way down to the Scanner.Scan to fill GitHeadRev. If we still want to support `acb build https://github.com/blah.git`, we will likely need a new ObtainSourceCode method to do something like the following:
```
retrieve the source code
if commit is provided, return commit
else try to git rev-parse to get the commit
```

cmd/acb/commands/build/build_test.go
```diff
@@ -51,7 +51,8 @@ func TestCreateBuildTask(t *testing.T) {
 		debug,
 		registry,
 		push,
-		creds)
+		creds,
+		"")
```
northtyphoon: maybe use a variable like `workingDirectory` to hold the empty string

scan/git.go
shahzzzam: i dont see why calling it 2x would be a problem..

shahzzzam: its like: `download` -> ObtainSourceCode (1st time, remote ctx) -> get commitID (inside workDir) -> pass it to builder

RunStep -> Scan -> ObtainSourceCode (2nd time, local ctx) -> get commitID (inside workDir)

PR #438, opened 2019-04-18
Update scanner to allow default Dockerfile path

### LINKED ISSUE #434: Find dockerfile within the context folder if not specified
BUG REPORT
**What happened?**:

TASK.YAML
- build -t blah some/other/folder 

It will try to find dockerfile in /workspace or c:\workspace

But it should try to find it in /workspace/some/other/folder 
**Environment**:

ehotinger:
**Purpose of the PR:**

In the case of `docker build src/foo`, the Dockerfile will be looked at `src/foo`. Right now since we always default the Dockerfile to `Dockerfile`, we check at `.` when scanning for dependencies. This change makes it so that if we default the Dockerfile, we'll check for it relative to the specified context / working directory.

You can view an example of the current issue as this yaml: https://github.com/ehotinger/scratch/blob/master/acb-nested.yaml -- without this change the dependencies will be chosen from the Dockerfile in the root of the repository; with the change the dependencies are looked at in the proper subfolder.

**Fixes #434**

### DIFF
--- a/scan/dependencies.go
+++ b/scan/dependencies.go
@@ -21,6 +21,7 @@ import (
 
 const (
 	dockerfileComment = "#"
+	defaultDockerfile = "Dockerfile"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


scan/dependencies.go
```diff
@@ -34,7 +34,9 @@ func (s *Scanner) ScanForDependencies(context string, workingDir string, dockerf
 	// In the case of remote contexts, it's scoped to the clone or downloaded location and the working directory.
 	// For example, for a git URL ending with .git#:foo/bar which was downloaded to a directory "build",
 	// the path is scoped to build/foo/bar/Dockerfile.
-	if !util.IsLocalContext(s.context) {
+	// If the Dockerfile was not specified, we default its value to "Dockerfile" and check it relative
+	// to the context.
+	if s.defaultedDockerfile || !util.IsLocalContext(s.context) {
```
northtyphoon: >defaultedDockerfile [](start = 6, length = 19)

If you just pass empty dockerfile to scanner, you can check here and assign dockerfilePath to `path.Clean(path.Join(workingDir, "Dockerfile"))`

Therefore you can avoid introducing the new `defaultedDockerfile` parameter.

PR #440, opened 2019-04-19
Refactor akv to keyvault

ehotinger:
**Purpose of the PR:**

- Refactor the `akv` yaml property to `keyvault`, this also renames many of the `Akv*` structs to be `KeyVault` instead.

### DIFF
--- a/docs/task-with-secrets.md
+++ b/docs/task-with-secrets.md
@@ -123,7 +123,7 @@ vim task-secrets.yaml
 version: 1.0-preview-1
 secrets:
   - id: mysecret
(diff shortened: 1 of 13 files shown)


docs/task-with-secrets.md
```diff
@@ -156,17 +156,17 @@ vim task-secrets.yaml
 version: 1.0-preview-1
```
northtyphoon: Can you replace `1.0-preview-1` in all docs and exampels?

ehotinger: Good catch, fixed

PR #443, opened 2019-05-07
Fix Task level environment variables

ehotinger:
**Purpose of the PR:**

- When default environment variables were introduced with MSI, these environment variables were overriding the entire Task's environment variables. I.e., if you had `env: [foo=bar]` at the Task level with no defaults, you'll end up with an empty slice. Now, we merge them together.

This also refactors some of the tests and functionality in that area as I was poking around.

### DIFF
--- a/builder/context_test.go
+++ b/builder/context_test.go
@@ -79,7 +79,7 @@ func TestGetImageDependencies(t *testing.T) {
 
 func TestGetBuildDockerRunArgs(t *testing.T) {
 	builder := &Builder{}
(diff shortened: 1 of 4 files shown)


northtyphoon:
> func mergeEnvs(stepEnvs []string, taskEnvs []string) ([]string, error) {

nit: maybe rename to dest, src

---
Refers to: graph/task.go:294 in a358be1. [](commit_id = a358be1fc36062c872e8d7ec37148b29ee594c98, deletion_comment = False)

PR #445, opened 2019-05-17
replace deadline exceeded message with timed out

### LINKED ISSUE #382: Refine the error message if the step timeout
FEATURE REQUEST

**What happened?**:

If the step failed as timeout, acb currently dumps the golang context error in the log. Eg,
It's a little hard to interpret the error message for people who are not familiar with the message.

```
Container failed during run: acb_step_0. No retries remaining.
Error: failed to run step ID: acb_step_0: context deadline exceeded
```

Provide a clear error message to tell users the step timeout after X duration.

northtyphoon:
The change add timeout information to the step logs.

2019/05/17 22:01:20 Executing step ID: acb_step_0. **Timeout(sec): 10**, Working directory: '', Network: 'acb_default_network'
2019/05/17 22:01:20 Launching container with name: acb_step_0
2019/05/17 22:01:30 Container failed during run: acb_step_0. No retries remaining.
failed to run step ID: acb_step_0: **timed out**

**Fixes #382**

### DIFF
--- a/pkg/procmanager/procmanager.go
+++ b/pkg/procmanager/procmanager.go
@@ -6,6 +6,7 @@ package procmanager
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"log"
 	"os"
@@ -33,7 +34,7 @@ func NewProcManager(dryRun bool) *ProcManager {
 }
 
 // RunRepeatWithRetries performs a Run multiple times with retries.
-// If an error occurs during the repetition, the last error will be returned.
+// If any error occurs during the repetition, all errors will be aggregated and returned.
 func (pm *ProcManager) RunRepeatWithRetries(
 	ctx context.Context,
 	args []string,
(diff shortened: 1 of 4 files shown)


pkg/procmanager/procmanager.go
```diff
@@ -33,7 +34,7 @@ func NewProcManager(dryRun bool) *ProcManager {
 }
 
 // RunRepeatWithRetries performs a Run multiple times with retries.
-// If an error occurs during the repetition, the last error will be returned.
+// If any error occurs during the repetition, all errors will be aggregated and returned.
```
sajayantony: So is this comment incorrect and we always returned all errors?

northtyphoon: yes, `aggErrors = append(aggErrors, innerErr)`

northtyphoon:
> I would try to handle it at the summary and check all errors’ cause for timeout

@ehotinger  "At summary" do you mean the utility function `Errors.Strings`?

northtyphoon:
OK. I did the simple string replace in main function (the ultimate code to print all errors). For the aggregated error in repeated step case, we need a new solution to `warp` the error from all step if we need to track all the original causing errors. /cc: @ehotinger

PR #465, opened 2019-06-25
Adds support for .Run.OSVersion.

ehotinger:
Signed-off-by: Eric Hotinger <ehotinger@gmail.com>

**Purpose of the PR**

- Adds support for .Run.OSVersion.


ehotinger:
We should refactor all the CLI commands at this point since there's a lot of duplicated code between all these shared options. #466 for that

shahzzzam:
where is OSVersion actually used? in docker?

ehotinger:
@shahzzzam will allow people using Windows to run / build images differently depending on our host

Issue #464, opened 2019-06-20
Proposed Alias Additions to Task Yaml


estebanreyl:
# Task Yaml alias additions

We have been discussing how to facilitate and improve on existing task yaml capabilities. In particular by looking at declaring aliases within the yaml files for reusability and general convenience. In current consideration we want something providing functionality similar to C's #define preprocessor directive, with some added changes, in particular:

    Through a global alias OSS file
    - More succinctly express common patterns. Eg. using $commit instead of {{.Run.Commit}}
    - Provide aliases for pre-defined well known tools. Eg. using $curl instead of mcr.microsoft.com/acr-task-commands/curl:latest
        - Containers would be pulled if not present locally on execution.
        - This would also allow easier integration with acr's other acr tools like Gil's new prune feature
    For this in particular we could allow distinct global definition files for users to choose from 
    outside of the default 

    # Using local files 
    - Provide definition only files (for reusability) to be included in task yaml files in a similar
        fashion to including an import.
    
    # Using local aliases
    - Users may define new values or override global ones in the definition section.

Defined values would be accessed using the $ character as a preceding char. $$ used to escape $.
All alias could also be disabled in the define portion of the yaml or the alias identifier $
can be re defined.

Alias definition should be hierarchical, such that the last override of an alias will be the final 
considered value, that is hierarchically speaking:
    local definitions > file definitions > Global definitions

As an example we could see:

Example global definitions:
```yaml
    registry : {{.Run.Registry}}
    commit : {{.Run.Commit}}
    id : {{.Run.ID}}
```

Example task yaml file:
```yaml
version: v1.0.0
    alias-src: # Import common definitions from external files (Read top down, further down indicates higher precedence)
        - org-defaults.yaml #Example of an organization default
        - proj-defaults.yaml #Additional definitions used by other local projects
    alias: # Local definitions (Take precedent over those from provided files) 
        - helm: cmd.azurecr.io/helm:v2.11.0-rc.2
    steps:
    # build website and func-test images, concurrently
    - build: -t $registry/hello-world:$id -f hello-world.dockerfile .
        when: ["-"]
    - build: -t hello-world-test  -f hello-world.dockerfile  .
        when: ["-"]
    # run built images to be tested
    - id: hello-world
        cmd: $registry/hello-world:$id
    - id: func-tests
        cmd: hello-world-test
        env:
        - test-target=hello-world
    # push hello-world if func-tests are successful  
    - push: 
        - $registry/hello-world:$id
[...]

PR #468, opened 2019-07-17
Alias Work

estebanreyl:
We have been discussing how to facilitate and improve on existing task yaml capabilities. In particular by looking at declaring aliases within the yaml files for reusability and general convenience. There's an ongoing high level discussion about this here: https://github.com/Azure/acr-builder/issues/464 . This draft implements the various proposed features although not all practices may be perfect (Golang and I are still only distant acquaintances).

I would appreciate some feedback on this work and will continue to improve upon it. As a note I have validated that everything here does run and properly works at least within the small testing framework I setup for this purpose at https://github.com/estebanreyl/preprocessor-test to enable use of the debugger and faster startup. Nonetheless everything should also work in this repo.


graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+			continue
+		} else {
+			out.WriteRune(char)
+		}
+	}
+
+	return out.String(), nil
+}
+
+// PreprocessBytes Handles files or byte encoded data that can be parsed through pre processing
```
northtyphoon: handles

northtyphoon: where are the files?

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+	"strings"
+	"time"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+var (
+	errUnknownAlias              = errors.New("unknown Alias")
+	errImproperDirectiveOverload = errors.New("$ directive can only be overwritten by a single character")
+	errImproperKeyName           = errors.New("alias key names only support alphanumeric characters and _ , - , ! characters")
```
northtyphoon: how do you choose `!`?

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/* Parses out alias  definitions from a given bytes array and appends
+them to the PreTask. Note alias definitions already in preTask will
+not be overwritten even if present in the array. */
+func readAliasFromBytes(data []byte, preTask *PreTask) error {
+
+	aliasMap := &map[string]string{}
+
+	if err := yaml.Unmarshal(data, aliasMap); err != nil {
```
northtyphoon: how does `yaml.Unmarshal` handle duplicated alias? Does is return error?

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+
+var (
+	errUnknownAlias              = errors.New("unknown Alias")
+	errImproperDirectiveOverload = errors.New("$ directive can only be overwritten by a single character")
+	errImproperKeyName           = errors.New("alias key names only support alphanumeric characters and _ , - , ! characters")
+	directive                    = '$'
+)
+
+// PreTask intermediate step for processing before complete unmarshall
+type PreTask struct {
```
northtyphoon: It's not very clear to me what is PreTask from the name. Maybe just Alias?

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+	directive rune
+}
+
+// Prevents recursive definitions from occuring
+func (preTask *PreTask) resolveMapAndValidate() error {
+	//Set directive from Map
+	preTask.directive = directive
+	if _, ok := preTask.AliasMap[string(directive)]; ok {
+		if len(preTask.AliasMap[string(directive)]) != 1 {
+			return errImproperDirectiveOverload
```
northtyphoon: If you allow override the directive, should it avoid the chars allowed for the alias name?

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	var out strings.Builder
+	var command strings.Builder
+	ongoingCmd := false
+
+	// Search and Replace
+	for _, char := range str {
+		if ongoingCmd {
+			//Maybe just checking if non alphanumeric, only allow alpha numeric aliases?
+			if strings.Contains(")}/ .,;]&|'~\n\t", string(char)) { // Delineates the end of an alias
```
northtyphoon: seems you need to check it is not in the allowed alias name charset

northtyphoon:
> I setup for this purpose at https://github.com/estebanreyl/preprocessor-test to enable use of the debugger and faster startup. Nonetheless everything should also work in this repo.

I assume you plan to migrate to the unit test here, right?

graph/preprocessor.go
estebanreyl: That doesn't have a good reason. I was mostly thinking as programmers it seems fun to have names like !home but that's probably bad practice anyway. I will remove that one.

graph/preprocessor.go
estebanreyl: Thats misslabeling on my side

graph/preprocessor.go
estebanreyl: I am not sure what you mean by this. I am also uncertain if allowing the override of the directive is even a good idea. I provide a escape sequence for it anyway.

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+	if err := yaml.Unmarshal(data, preTask); err != nil {
+		return nil, err
+	}
+
+	if preTask.AliasMap == nil && preTask.AliasSrc == nil {
+		return data, nil
+	}
+
+	// Search and Replace
+	str := string(data[:])
```
ehotinger: `[:]` can be removed

graph/preprocessor.go
ehotinger: could you name all the functions and structs lowercase since they are private to this package and there's no reason to export them (yet)

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+			out.WriteRune(char)
+		}
+	}
+
+	return out.String(), nil
+}
+
+// PreprocessBytes Handles files or byte encoded data that can be parsed through pre processing
+func PreprocessBytes(data []byte) ([]byte, error) {
+	preTask := &PreTask{}
```
ehotinger: an optimization here - if there are no aliases defined then there's no reason to replace anything or unmarshal it twice - since the end result will be the same. So it might make sense to just unmarshal the Task as a Task normally.

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+* files. This is done by unmarshaling these elements which will then be added in a hierarchical
+* manner. Note the input must still be valid Yaml.
+*
+* Existing issues: Once the preTask is parsed out and the definitions are resolved, the read in
+* yaml will be processed to include the appropriate aliases, however this will include the previously
+* parsed in definitions which is not as efficient.
+*
+ */
+
+package graph
```
ehotinger: Could you add a TODO for the default global alias mapping, i.e. `$az => azure-cli:someTag` - this is a p0 on the initial implementation

graph/preprocessor.go
```diff
@@ -0,0 +1,211 @@
...
+var (
+	errUnknownAlias              = errors.New("unknown Alias")
+	errImproperDirectiveOverload = errors.New("$ directive can only be overwritten by a single character")
+	errImproperKeyName           = errors.New("alias key names only support alphanumeric characters and _ , - , ! characters")
+	directive                    = '$'
+)
+
+// PreTask intermediate step for processing before complete unmarshall
+type PreTask struct {
+	AliasSrc  []*string         `yaml:"alias-src"`
```
ehotinger: Be sure to run the names of the yaml through Steve. Bare minimum we don't have dashes anywhere and use camelCase instead so it would be good to change that.

graph/preprocessor.go
ehotinger: 100% unit test coverage for everything in this file 👍

graph/preprocessor.go
estebanreyl: I will make sure to do that

graph/preprocessor.go
estebanreyl: I will look at how to get this done

graph/preprocessor.go
estebanreyl: Yes, it might also be good to meet to see which global ones we should be providing since I have less context on what images people are usually using for example

graph/preprocessor.go
estebanreyl: That... Is actually a much better idea than what I am doing. Thanks Bin!

graph/preprocessor.go
estebanreyl: That is true and I've given it some thought, nonetheless I am not sure if there is a way to prevent the double unmarshaling as I need to first confirm there are no aliases defined. The alternative would be to unmarshall the whole thing (as in no PreTask struct and just an extended task struct) and then modifying this resulting structure but that would require additional parsing of the resulting dynamic map and I am not sure that's much better. Do you have any thoughts about how I may approach this?

shahzzzam:
@estebanreyl you can move PR from Draft to "Ready for review". Also, once you make changes - check the CI (acr-builder (Linux) & acr-builder (Windows)) to see why they fail.

northtyphoon:
There are some lint errors in CI build.

https://dev.azure.com/azurecontainerregistry/acr-builder/_build/results?buildId=412

shahzzzam:
Make sure to rebase your changes from upstream and resolve the CI errors.

graph/preprocessor.go
```diff
@@ -0,0 +1,281 @@
...
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+var (
+	errImproperDirectiveLength = errors.New("$ directive can only be overwritten by a single character")
+	errImproperKeyName         = errors.New("alias key names only support alphanumeric characters")
+	errImproperDirectiveChoice = errors.New("overwritten directives may not be alphanumeric characters")
+	directive                  = '$'
+	aliasFormat                = regexp.MustCompile("\\A[a-zA-Z0-9]+\\z")
```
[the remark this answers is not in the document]
estebanreyl: No, I mean this is generally faster. I meant for when I am validating the aliases themselves (making sure they are alphanumeric by using the regex instead of looping over each character and checking this?)

cmd/acb/commands/exec/exec.go
```diff
@@ -194,6 +196,32 @@ var Command = cli.Command{
...
+		type Wrapper struct {
+			Version string `yaml:"version,omitempty"`
+		}
+
+		wrap := &Wrapper{}
+		if err := yaml.Unmarshal(template.GetData(), wrap); err != nil {
+			return err
+		}
+
+		shouldIncludeAlias := wrap == nil || wrap.Version >= "v.1.1.0"
```
shahzzzam: `wrap` will never be `nil`. Try `wrap.Version == ""` instead

shahzzzam: nit : `v1.1.0`

There's an extra dot

cmd/acb/commands/exec/exec.go
```diff
@@ -194,6 +196,32 @@ var Command = cli.Command{
...
+		var task *graph.Task
+		if shouldIncludeAlias {
+			// Generate the base task file without resolving environment variables.
+			task, err := graph.NewTaskFromBytes(template.GetData())
+			if err != nil {
+				return err
+			}
+
+			// Remarshall functional task file to resolve templating
```
shahzzzam: Remarshal* ?

cmd/acb/commands/exec/exec.go
```diff
@@ -215,9 +243,20 @@ var Command = cli.Command{
...
-		if err != nil {
-			return err
+		// If work has been done before as a result of preprocessing avoid re
+		// computing the TaskFrom a string.
+		if shouldIncludeAlias {
+			task.CompleteTask(ctx, defaultWorkingDirectory, defaultNetwork, defaultEnvs, credentials)
+			if err != nil {
+				return err
+			}
+
```
shahzzzam: extra line

cmd/acb/commands/exec/exec.go
```diff
@@ -194,6 +196,32 @@ var Command = cli.Command{
...
+			// Remarshall functional task file to resolve templating
+			fromTask, err := yaml.Marshal(task)
+			if err != nil {
+				return err
+			}
+			template.Data = fromTask
+		}
+
 		rendered, err := templating.LoadAndRenderSteps(ctx, template, renderOpts)
```
shahzzzam: This `rendered` string will have rendered things like: 
{{.Run.Registry}}

Your code never takes this path (i.e. uses rendered to to UnmarshalTaskFromString.

That is a bug. Shouldnt it be in this sequence:

1. Load up all the "Task aliases" in memory (in the `aliasMap`.)
For instance, 
> commit: "{{.Run.Commit}}"
repository: "{{.Run.Repository}}"
branch: "{{.Run.Branch}}"
gitTag: "{{.Run.GitTag}}"
triggeredBy: "{{.Run.TriggeredBy}}"
registry: "{{.Run.Registry}}"
registryName: "{{.Run.RegistryName}}"
date: "{{.Run.Date}}"
sharedVolume: "{{.Run.SharedVolume}}"
os: "{{.Run.OS}}"
architecture: "{{.Run.Architecture}}"
id: "{{.Run.ID}}"
osArchitecture: ""
//  Usable without $
purge: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e purge"
tagList: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e tag list"
manifestList: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e manifest list"
manifestDelete: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e manifest delete"
tagDelete: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e tag delete"
// Usable without directive
acr: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e"
az: "mcr.microsoft.com/azure-cli"
bash: "bash" # Note for this one bash is in fact the alias and the source at dockerhub
buildkit:
buildx: ""
curl: "appropriate/curl"
docker: "docker" # Note for this one docker is in fact the alias and the source at dockerhub
dotnet: "mcr.microsoft.com/dotnet/core"
git: ""
go: "golang"
helm: "cmd.azurecr.io/helm:v2.11.0-rc.2"
kubectl: "bitnami/kubectl"
npm: "chekote/node"
wget: "abihf/wget"
yarn: "chekote/node"

And some more...

2) Preprocess the string Task
3) "LoadAndRenderSteps" aka Render any go templating in the string 
4) UnmarshalTaskFromString.

I suggest separating out the "preprocessing" logic from "NewTaskFromBytes". IMO, the Task object should only be made after Rendering is done.

What do you say, @northtyphoon ?

graph/preprocessor.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	var out strings.Builder
+	var command strings.Builder
+	ongoingCmd := false
+	changed := false
+
+	// Search and replace all strings with the directive
+	for _, char := range str {
+		if ongoingCmd {
+			if char == alias.directive && command.Len() == 0 { // Escape Character Triggered
```
shahzzzam: Can you point me to the code which takes care of 
```
- cmd: bash ...
```
i.e. the case without directive?

cmd/acb/commands/exec/exec.go
estebanreyl: Yes but there is an issue with that. The only reason why I had to do this was as a result of the directive-less image names under cmd : In order to process those properly we must have a task already read into a task object so we can determine the location of cmd values. As a result, in order to avoid rendering twice I first fully obtain the task file.

graph/global-defaults.yaml
```diff
@@ -0,0 +1,39 @@
...
+gitTag: "{{.Run.GitTag}}"
+triggeredBy: "{{.Run.TriggeredBy}}"
+registry: "{{.Run.Registry}}"
+registryName: "{{.Run.RegistryName}}"
+date: "{{.Run.Date}}"
+sharedVolume: "{{.Run.SharedVolume}}"
+os: "{{.Run.OS}}"
+architecture: "{{.Run.Architecture}}"
+id: "{{.Run.ID}}"
+osArchitecture: ""
```
northtyphoon: Please reference to [https://github.com/Azure/acr-builder/blob/master/docs/templates.md#run-variables](https://github.com/Azure/acr-builder/blob/master/docs/templates.md#run-variables)

graph/preprocessor.go
estebanreyl: check the func processSteps(alias *Alias, task *Task). Its called at the end of newTaskFromBytes

graph/global-defaults.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+# Usable without $
+# Containerized commands can be used without $ directive
+purge: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e purge"
```
northtyphoon: >mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e [](start = 8, length = 43)

update the image to mcr.microsoft.com/acr/acr-cli:0.1

graph/global-defaults.yaml
estebanreyl: Ill match the names put forth there

graph/global-defaults.yaml
```diff
@@ -0,0 +1,39 @@
...
+curl: "appropriate/curl"
+docker: "docker" # Note for this one docker is in fact the alias and the source at dockerhub
+dotnet: "mcr.microsoft.com/dotnet/core"
+git: ""
+go: "golang"
+helm: "cmd.azurecr.io/helm:v2.11.0-rc.2"
+kubectl: "bitnami/kubectl"
+npm: "chekote/node"
+wget: "abihf/wget"
+yarn: "chekote/node"
```
northtyphoon: I would recommend to start from a small set initially.

acr
az
curl
bash
kubectl
helm
git

graph/global-defaults.yaml
```diff
@@ -0,0 +1,39 @@
...
+id: "{{.Run.ID}}"
+osArchitecture: ""
+
+# Usable without $
+# Containerized commands can be used without $ directive
+purge: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e purge"
+tagList: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e tag list"
+manifestList: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e manifest list"
+manifestDelete: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e manifest delete"
+tagDelete: "mcr.microsoft.com/acr/acr-cli:linux-7ccdc3e tag delete"
```
northtyphoon: are these command still valid?

[https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auto-purge](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auto-purge)

graph/global-defaults.yaml
```diff
@@ -0,0 +1,39 @@
+# Convenience definitions, use as eg. $commit
```
northtyphoon: We need sperate files for windows and linux

graph/global-defaults.yaml
estebanreyl: I think they were removed, will update accordingly

graph/global-defaults.yaml
estebanreyl: I've restricted the set, do we know how to acquire the endpoints at this point however? I am not sure if/where we may have images for curl, bash, kubectl, helm or git that we may want to use.

graph/global-defaults.yaml
estebanreyl: How do we determine what kind of environment the task will be operating in?/Are most  of these images not multi-arch?

graph/preprocessor.go
```diff
@@ -0,0 +1,291 @@
...
+
+	return []byte(parsedStr), *alias, changed, err
+}
+
+// processSteps Will resolve image names in steps that are aliased without using $.
+// Invoked after resolving $
+func processSteps(alias *Alias, task *Task) {
+	for i, step := range task.Steps {
+		parts := strings.Split(step.Cmd, " ")
+		if _, ok := alias.AliasMap[parts[0]]; ok {
```
shahzzzam: `step.Cmd` could be empty. You will get index out of range for `parts[0]`

graph/preprocessor.go
```diff
@@ -0,0 +1,291 @@
...
+	parsedStr, changed, err := preprocessString(alias, str)
+
+	return []byte(parsedStr), *alias, changed, err
+}
+
+// processSteps Will resolve image names in steps that are aliased without using $.
+// Invoked after resolving $
+func processSteps(alias *Alias, task *Task) {
+	for i, step := range task.Steps {
+		parts := strings.Split(step.Cmd, " ")
```
shahzzzam: `if step.Cmd != ""`  {

}

graph/preprocessor.go
```diff
@@ -0,0 +1,291 @@
...
+		if _, ok := alias.AliasMap[parts[0]]; ok {
+			// Image name should always go first
+			parts[0] = alias.AliasMap[parts[0]]
```
shahzzzam: Cannot assume every command will have an alias as the first argument. For the following command:

```
- cmd: mcr.microsoft.com/acb exec ...
```

`alias.AliasMap["mcr.microsoft.com/acb"] does not exist.

```
if val, ok := alias.AliasMap[parts[0]]; ok {
    parts[0] = alias.AliasMap[parts[0]]
    task.Steps[i].Cmd = strings.Join(parts, " ")
}
```

graph/preprocessor.go
estebanreyl: That wont crash. If Cmd is empty (Since it'll be an empty string) Split will return an array with only the empty string in it at position 0. As a result this will work and because aliases cannot be the empty string the if will never return ok = true

graph/preprocessor.go
estebanreyl: This shouldn't be necessary as above but may be useful if the readability of it is confusing

graph/preprocessor.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// preprocessBytes handles byte encoded data that can be parsed through pre processing
+func preprocessBytes(data []byte) ([]byte, Alias, bool, error) {
+	type Wrapper struct {
+		Alias Alias `yaml:"alias,omitempty"`
+	}
+	wrap := &Wrapper{}
+	aliasData, remainingData, err := basicAliasSeparation(data)
+	if errUnMarshal := yaml.Unmarshal(aliasData, wrap); errUnMarshal != nil {
```
shahzzzam: nit: errUnmarshal*

graph/preprocessor.go
estebanreyl: I am a little confused by the first assumption. From my understanding, I am only replacing the value if the first argument can be detected to be an alias through its presence in the map. Are you speaking about the assumption that it will be the first item? Eric told me that was a fair assumption as while this may not always be true, its rare for people to do something different and we wish to encourage the behavior where the image name comes first (or an alias in this scenario). If the value has no alias , in this case, its just kept as it was before.

graph/task.go
```diff
@@ -66,6 +66,12 @@ func UnmarshalTaskFromString(ctx context.Context, data string, defaultWorkDir st
 	if err != nil {
 		return t, errors.Wrap(err, "failed to deserialize task and validate")
 	}
+	t.CompleteTask(ctx, defaultWorkDir, network, envs, creds)
+	return t, err
+}
+
+// CompleteTask unmarshals a Task from a raw string.
```
shahzzzam: CompleteTask prepares a Task with remaining parameters

graph/task.go
```diff
@@ -66,6 +66,12 @@ func UnmarshalTaskFromString(ctx context.Context, data string, defaultWorkDir st
 	if err != nil {
 		return t, errors.Wrap(err, "failed to deserialize task and validate")
 	}
+	t.CompleteTask(ctx, defaultWorkDir, network, envs, creds)
```
shahzzzam: Grab the error from complete task and return if err

graph/global-defaults-linux.yaml
```diff
@@ -0,0 +1,18 @@
+# Convenience definitions, use as eg. $commit
```
shahzzzam: nit: $Commit*

shahzzzam: "Native Task aliases"

graph/util.go
```diff
@@ -0,0 +1,21 @@
...
+func getTestGlobalDefaults() string {
+
+	var fileLoc string
+	if runtime.GOOS == "windows" {
+		fileLoc = "./global-defaults-windows.yaml"
+	} else { // Looking at Linux
+		fileLoc = "./global-defaults-linux.yaml"
+	}
+	return fileLoc
+}
```
northtyphoon: Can you check `util/sock_linux.go` and `util/sock_windows.go` to use platform-aware const variable?

cmd/acb/commands/exec/exec.go
```diff
@@ -221,9 +274,21 @@ var Command = cli.Command{
...
 		}
 
-		task, err := graph.UnmarshalTaskFromString(ctx, rendered, defaultWorkingDirectory, defaultNetwork, defaultEnvs, credentials, taskName)
-		if err != nil {
-			return err
+		// If work has been done before as a result of preprocessing avoid re
+		// computing the TaskFrom a string.
+		if shouldIncludeAlias {
+			log.Printf("(exec) task completion started")
+			err := task.CompleteTask(ctx, defaultWorkingDirectory, defaultNetwork, defaultEnvs, credentials, taskName)
```
shahzzzam: Can you rename the method to something else? CompleteTasks suggests that the task will complete/run in this method.

graph/util.go
estebanreyl: Should I be checking these paths ?. Not fully certain what these do

Dockerfile
```diff
@@ -8,16 +8,18 @@ RUN apk add --no-cache \
 FROM gobuild-base AS acb
 WORKDIR /go/src/github.com/Azure/acr-builder
 COPY . .
-RUN make binaries && mv bin/acb /usr/bin/acb
+RUN mkdir -p /usr/bin/builder-files
+COPY ./graph/global-defaults-linux.yaml /usr/bin/builder-files/global-defaults-linux.yaml
```
shahzzzam: revert this dockerfile change

Windows.Dockerfile
```diff
@@ -88,24 +88,26 @@ FROM builder as dockercli
...
 # Build the acr-builder
 FROM builder as acb
 COPY --from=dockercli /gopath/src/github.com/docker/cli/build/docker.exe c:/docker/docker.exe
 WORKDIR \\gopath\\src\\github.com\\Azure\\acr-builder
 COPY ./ /gopath/src/github.com/Azure/acr-builder
+RUN mkdir -p builder-files
+COPY .\graph\global-defaults-windows.yaml builderfiles/
 RUN Write-Host ('Running build'); \
-    go build -o acb.exe .\cmd\acb; \
+	go build -o builder-files\acb.exe .\cmd\acb; \
```
shahzzzam: revert this dockerfile change

cmd/acb/commands/exec/exec.go
```diff
@@ -200,6 +205,54 @@ var Command = cli.Command{
...
+		if err = yaml.Unmarshal(template.GetData(), wrap); err != nil {
+			return err
+		}
+
+		shouldIncludeAlias := wrap.Version == "" || wrap.Version >= "v1.1.0"
+		var task *graph.Task
+		if shouldIncludeAlias {
+			log.Printf("(exec) alias support detected")
+
+			//Identify defaults location.
```
shahzzzam: nit: space after //

cmd/acb/commands/exec/exec.go
```diff
@@ -200,6 +205,54 @@ var Command = cli.Command{
...
+			var ex string
+			ex, err = os.Executable()
```
shahzzzam: remove this code

graph/global-defaults-linux.yaml
```diff
@@ -0,0 +1,20 @@
+# Native Task aliases, use as eg. $Commit
+ID: "{{.Run.ID}}"
+SharedVolume: "{{.Run.SharedVolume}}"
+Registry: "{{.Run.Registry}}"
+RegistryName: "{{.Run.RegistryName}}"
```
shahzzzam: remove these files global defaults files

graph/util.go
shahzzzam: remove loading aliases from yaml files.

Windows.Dockerfile
shahzzzam: revert both dockerfile changes

Dockerfile
```diff
@@ -1,3 +1,4 @@
+
```
shahzzzam: remove line. Infact, remove these two Dockerfiles from your PR altogether.

graph/preprocessor.go
```diff
@@ -0,0 +1,310 @@
...
+	return nil
+}
+
+// Loads in all Aliases defined as being a part of external resources.
+func (alias *Alias) loadExternalAlias() error {
+
+	// Iterating in reverse to easily and efficiently handle hierarchy. The later
+	// declared the higher in the hierarchy of alias definitions.
+	for i := len(alias.AliasSrc) - 1; i >= 0; i-- {
+
```
shahzzzam: nit: if possible, please clean up all these additional lines in all the files

graph/preprocessor.go
```diff
@@ -0,0 +1,281 @@
...
+		parts := strings.Split(step.Cmd, " ")
+		if _, ok := alias.AliasMap[parts[0]]; ok {
+			// Image name should always go first
+			parts[0] = alias.AliasMap[parts[0]]
+			task.Steps[i].Cmd = strings.Join(parts, " ")
+		}
+	}
+}
+
+// Provides simple separation of the top level items in a yaml file definition, however, need
```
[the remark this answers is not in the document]
shahzzzam: remove this compliance part from the comment. Update the comment

graph/preprocessor.go
```diff
@@ -0,0 +1,320 @@
...
+			return errImproperKeyName
+		}
+	}
+
+	return nil
+}
+
+// Loads in all Aliases defined as being a part of external resources.
+func (alias *Alias) loadExternalAlias() error {
+
```
shahzzzam: nit: extra line

graph/preprocessor.go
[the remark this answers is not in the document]
shahzzzam: @estebanreyl update the comment

graph/preprocessor.go
```diff
@@ -0,0 +1,320 @@
...
+			aliasBuffer.WriteString(text + "\n")
+		} else {
+			buffer.WriteString(text + "\n")
+		}
+	}
+
+	return aliasBuffer.Bytes(), buffer.Bytes()
+}
+
+// FindVersion Determines the current version of an Alias task file
```
shahzzzam: nit: determines

graph/preprocessor.go
```diff
@@ -0,0 +1,320 @@
...
+func FindVersion(data []byte) string {
+	reader := bytes.NewReader(data)
+	scanner := bufio.NewScanner(reader)
+	scanner.Split(bufio.ScanLines)
```
shahzzzam: no need. its the default:
https://golang.org/pkg/bufio/#Scanner.Split

remove it

graph/preprocessor.go
```diff
@@ -0,0 +1,320 @@
...
+	commentRe := regexp.MustCompile(`\A\s*#.*`)
+	versionField := regexp.MustCompile(`\Aversion\s*:.+\z`)
+
+	for scanner.Scan() {
+		text := scanner.Text()
+
+		//Finds the first non empty and non comment string
+		comment := commentRe.MatchString(text)
```
shahzzzam: Have you considered whitespace? In your case a line containing whitespace will `break`. 

Remove usage of comments and version regex. Do this instead:

```go
for scanner.Scan()  {
  text := scanner.Text()

  trimmedText = strings.TrimSpace(text)

  // Skip comments and just whitespace.
  if trimmedText == "" || strings.HasPrefix(trimmedText, "#") {
    continue
  }

  // use text instead of trimmedText since '   version: ' is also invalid. 
  if strings.HasPrefix(text, "version:") {
    return strings.TrimPrefix(trimmedText, "version:")
  }
  break
}
```

graph/preprocessor.go
estebanreyl: May be the default but removing bufio.ScanLines results in build not working. I am gonna keep it then.

graph/preprocessor.go
estebanreyl: Nice solution, I like it

graph/preprocessor.go
shahzzzam: really? thats odd. We use the similar logic to parse Dockerfiles to build them.

https://github.com/Azure/acr-builder/blob/master/scan/dependencies.go#L164

We also parse that line by line , ignore whitespaces, comments, etc. 

But anyway. We can look at it later. You can keep it for now

PR #469, opened 2019-07-19
support buildcache

shahzzzam:
**Purpose of the PR**

- 

Fixes #


graph/step.go
```diff
@@ -12,6 +12,8 @@ import (
 
 	"github.com/Azure/acr-builder/pkg/image"
 	"github.com/Azure/acr-builder/util"
+	"github.com/docker/distribution/reference"
+	pkgerrors "github.com/pkg/errors"
```
ehotinger: remove `pkgerrors` and remove the other `errors` import. Only one and it should cover `.New` and `.Wrap`

graph/step.go
```diff
@@ -196,3 +202,42 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+func (s *Step) UseBuildCacheForBuildStep() bool {
+	if s.EnableCache && s.CacheID == "" {
+		s.CacheID = "cache"
+	}
+
+	return s.IsBuildStep() && s.CacheID != ""
+}
+
+// FetchBuildWithBuildCache returns an updated build step by adding buildx cache parameters
+func (s *Step) FetchBuildWithBuildCache() (string, error) {
```
ehotinger: Method name could be improved since it doesn't do anything related to its description

graph/step.go
```diff
@@ -196,3 +202,42 @@ func (s *Step) UpdateBuildStepWithDefaults() {
 		s.Build = fmt.Sprintf("--isolation hyperv %s", s.Build)
 	}
 }
+
+// UseBuildCacheForBuildStep indicates if buildx needs to be used
+func (s *Step) UseBuildCacheForBuildStep() bool {
```
ehotinger: Can you add unit tests for all of these

graph/step.go
```diff
@@ -196,3 +202,42 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+		return result, pkgerrors.New("(buildcache) did not provide any tags")
+	}
+
+	buildImg := s.Tags[0]
+	repo, err := reference.Parse(buildImg)
+	if err != nil {
+		return result, pkgerrors.Wrap(err, "(buildcache) failed to parse the buildimage")
+	}
+
+	// buildImg: "b.gcr.io/test.example.com/my-app:test.example.com",
```
ehotinger: delete L229 - L231

graph/step.go
```diff
@@ -196,3 +202,42 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+	// s.CacheID: "myBuildCache"
+	// cacheImage: "b.gcr.io/test.example.com/my-app:myBuildCache"
+	if named, ok := repo.(reference.Named); ok {
+		cacheImage, err := reference.WithTag(named, s.CacheID)
+		if err != nil {
+			return result, pkgerrors.Wrap(err, "(buildcache) failed to attach cacheID tag to the repo")
+		}
+
+		// todo (sam) check if we need `--push` to be able to push the cache.
```
ehotinger: Makes sense to me.

graph/step.go
```diff
@@ -196,3 +202,42 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+		return result, pkgerrors.Wrap(err, "(buildcache) failed to parse the buildimage")
+	}
+
+	// buildImg: "b.gcr.io/test.example.com/my-app:test.example.com",
+	// s.CacheID: "myBuildCache"
+	// cacheImage: "b.gcr.io/test.example.com/my-app:myBuildCache"
+	if named, ok := repo.(reference.Named); ok {
+		cacheImage, err := reference.WithTag(named, s.CacheID)
+		if err != nil {
+			return result, pkgerrors.Wrap(err, "(buildcache) failed to attach cacheID tag to the repo")
```
ehotinger: nit: the (buildcache) prefix is really uncommon and will look weird to users

graph/task.go
```diff
@@ -251,7 +251,9 @@ func (t *Task) initialize(ctx context.Context) error {
 		s.StepStatus = Skipped
 
 		if s.IsBuildStep() {
-			s.Tags = util.ParseTags(s.Build)
+			if s.Tags != nil && len(s.Tags) == 0 {
```
ehotinger: Why did you change this?

graph/step.go
```diff
@@ -60,6 +63,7 @@ type Step struct {
 	IgnoreErrors                    bool `yaml:"ignoreErrors"`
 	DisableWorkingDirectoryOverride bool `yaml:"disableWorkingDirectoryOverride"`
 	Pull                            bool `yaml:"pull"`
+	EnableCache                     bool `yaml:"enableCache"`
```
ehotinger: I wonder if it's really worth having this. Like if people don't want it can't they just update their YAML. I don't like the idea of adding more and more parameters to support upstream. I would prefer if they could do everything in YAML.

builder/builder.go
```diff
@@ -236,7 +241,24 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step) error {
...
+			if buildStepWithBuildCache, err := step.FetchBuildWithBuildCache(); err != nil {
+				log.Printf("error using buildcache %s\n", err)
+			} else {
+				useBuildCache = true
+				step.Build = buildStepWithBuildCache
+			}
+		}
+
+		if useBuildCache {
+			args = b.getDockerRunArgs(volName, workingDirectory, step, step.Envs, "", "buildx build "+step.Build)
```
ehotinger: can you turn the docker / buildx into img consts

builder/builder.go
```diff
@@ -236,7 +241,24 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step) error {
 			step.Build = replacePositionalContext(step.Build, ".")
 		}
 		step.UpdateBuildStepWithDefaults()
-		args = b.getDockerRunArgs(volName, workingDirectory, step, step.Envs, "", "docker build "+step.Build)
+		useBuildCache := false
+
+		// buildx is only supported on Linux
+		if runtime.GOOS == linuxOS && step.UseBuildCacheForBuildStep() {
+			log.Println("runtime os is linux and need to use buildcache")
+			if buildStepWithBuildCache, err := step.FetchBuildWithBuildCache(); err != nil {
```
ehotinger: is it easy to refactor this and put it in the Task processing/deserialization/initialization area instead? I never liked how the other one ended up here in the first place

builder/builder.go
```diff
@@ -236,7 +241,24 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step) error {
...
+				log.Printf("error using buildcache %s\n", err)
```
ehotinger: %v

builder/builder.go
```diff
@@ -236,7 +241,24 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step) error {
 			step.Build = replacePositionalContext(step.Build, ".")
 		}
 		step.UpdateBuildStepWithDefaults()
-		args = b.getDockerRunArgs(volName, workingDirectory, step, step.Envs, "", "docker build "+step.Build)
+		useBuildCache := false
+
+		// buildx is only supported on Linux
+		if runtime.GOOS == linuxOS && step.UseBuildCacheForBuildStep() {
```
ehotinger: might want to error or at least print a message if Windows.

graph/step.go
ehotinger: mmm, I don't know how I feel about `az acr build --enable-cache` just so that we can make a gesture to use buildx. I think `az acr build` should just run the default runtime and if they want complex stuff they need to move to YAML. So I question all the `cacheId`, `enableCache`, etc. stuff for upstream. It adds a lot of stuff to maintain for little value IMO

graph/step.go
ehotinger: makes sense to me

graph/step.go
shahzzzam: `GetCmdWithCacheFlags()` ?

graph/step.go
shahzzzam: I ll remove this once i have tested. I have added additional logging for testing purposes

graph/task.go
shahzzzam: Tha's because in the `build.go` we grab the tags and prefix them with registry (if necessary). I also added the list of those prefixedTags to the buildStep so that we don;'t have to go thru the command again.

```
// After the template has rendered, we have to parse the tags again
	// so we can properly set the build/push tags.
	rendered, prefixedTags := util.PrefixTags(rendered, registry)
	tags = prefixedTags

	buildStep := &graph.Step{
		ID:          "build",
		Build:       rendered,
		Timeout:     buildTimeoutInSec,
		EnableCache: enableCache,
		CacheID:     cacheID,
		Tags:        tags,    // <----
	}
```

graph/step.go
shahzzzam: talked offline to @northtyphoon .

I will remove them from `acb build` as we want things to go thru `exec` eventually. And I agree we should encourage advanced users to move to YAML anyway.

The design for exec currently is:
```
  - build: -t myreg.azurecr.io/repo:tag ... ..
    cacheId: "myCacheTag"

  - build: -t myreg.azurecr.io/repo:tag2
    enableCache: true
```

I am not fully sure why something like `{{.Run.PreviousRunCacheId}}` would be useful? Is there any such usecase?

Like do u mean something like this:
```
  - build: -t myreg.azurecr.io/repo:tag2
    cacheId: {{.Run.PreviousRunCacheId}}
```

That would also require us to fetch the previous run record.

//cc @northtyphoon @sajayantony

graph/task.go
ehotinger: and what about for exec

shahzzzam: Sorry, there was a bug in my code earlier. I fixed the condition now.

basically if the `s.Tags` is not set already, parseTags. 
When it goes thru build, they are already set - so we dont parseTags again.
When it goes thru exec, they are not - and we parseTags.

builder/builder.go
shahzzzam: https://github.com/Azure/acr-builder/issues/472

yep i will. I could just move the methods/calling during task.initialize() but, i then i will have to fix a couple of tests bc we add "--isolation hyperv" stuff for windows but not on linux.

graph/step.go
```diff
@@ -196,3 +201,49 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+
+// GetCmdWithCacheFlags adds buildx cache parameters to the cmd
+func (s *Step) GetCmdWithCacheFlags() (string, error) {
+	result := ""
+
+	if s == nil {
+		return result, errors.New("step is null")
+	}
+
+	if s.Tags == nil || len(s.Tags) == 0 {
```
shizhMSFT: nit: In golang, `len([]interface{}(nil))` is `0`

graph/step.go
```diff
@@ -196,3 +201,49 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+		return result, errors.New("atleast one tag is required to use buildcache's cache registry")
```
shizhMSFT: space?

graph/step.go
```diff
@@ -196,3 +201,49 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+
+	if named, ok := repo.(reference.Named); ok {
+		// also trim out any digests
+		named = reference.TrimNamed(named)
+		cacheImage, err := reference.WithTag(named, s.CacheID)
+		if err != nil {
+			return result, errors.Wrap(err, "failed to attach cache ID tag to the repo for buildcache")
+		}
+
+		// todo (sam) check if we need `--push` to be able to push the cache.
```
shizhMSFT: `--push` is not used for pushing the cache. It is used to push the resulted image.

Basically, `buildx` uses a `buildkitd` container to build the image as the build driver since you did `buildx create`. The container is a different system that it does not share images with the docker daemon outside. As a resolution, `buildx build` has an option [-o, --output](https://github.com/docker/buildx/blob/master/README.md#-o---outputpath-typetypekeyvalue) to export the resulted image. Two notable extended options are 
- `--push Shorthand for --output=type=registry` push the image after building completion
- `--load Shorthand for --output=type=docker` export the image to docker daemon after building completion so that you can find the image using `docker images` later.

graph/task.go
```diff
@@ -251,7 +251,9 @@ func (t *Task) initialize(ctx context.Context) error {
 		s.StepStatus = Skipped
 
 		if s.IsBuildStep() {
-			s.Tags = util.ParseTags(s.Build)
+			if s.Tags == nil || len(s.Tags) == 0 {
```
shizhMSFT: nit: `if len(s.Tags) == 0 {`

graph/step.go
shizhMSFT: You may take a look at `--load`. This option may be useful for Tasks.

builder/builder.go
```diff
@@ -236,7 +243,27 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step) error {
...
 		}
 		step.UpdateBuildStepWithDefaults()
-		args = b.getDockerRunArgs(volName, workingDirectory, step, step.Envs, "", "docker build "+step.Build)
+		useBuildCache := false
+
+		if step.UseBuildCacheForBuildStep() {
+			if runtime.GOOS == linuxOS {
+				log.Println("runtime os is linux and need to use buildcache")
+				if buildStepWithBuildCache, err := step.GetCmdWithCacheFlags(); err != nil {
+					log.Printf("error using buildcache %v\n", err)
```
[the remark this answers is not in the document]
shahzzzam: how to print `err` object?

graph/step.go
shahzzzam: Thanks, seems like `--push` is what I am looking for. Please see the updated code. How would `--load` be helpful?

builder/constants.go
```diff
@@ -22,4 +22,11 @@ const (
 	loginTimeoutInSec   = 60 * 5  // 5 minutes
 	digestsTimeoutInSec = 60 * 5  // 5 minutes
 	scrapeTimeoutInSec  = 60 * 15 // 15 minutes
+
+	// build cache constants
+	buildkitContainerRunTimeout     = 60 * 2 // 2 minutes
+	buildkitContainerInitRetries    = 3
+	buildkitContainerInitRetryDelay = 5 // 5 seconds
+	buildkitContainerName           = "buildkitcontainer"
```
northtyphoon: >buildkitcontainer [](start = 36, length = 17)

can you add a prefix like acr to the system container name?

builder/context.go
```diff
@@ -35,15 +35,17 @@ func (b *Builder) getDockerRunArgs(
...
 	var args []string
 	var sb strings.Builder
+	addWorkdir := true
+
 	// Run user commands from a shell instance in order to mirror the shell's field splitting algorithms,
 	// so we don't have to write our own argv parser for exec.Command.
 	if runtime.GOOS == windowsOS {
 		args = []string{"powershell.exe", "-Command"}
-		if step.Isolation == "" && !step.IsBuildStep() {
+		if step != nil && step.Isolation == "" && !step.IsBuildStep() {
```
northtyphoon: >step != nil [](start = 5, length = 11)

what is the new case that can pass null step to the function?

northtyphoon: I assume the function is only called to create run args for a step. If the step is null, it make me think if it's better to create a new function to handle the case instead of reusing the existing one. As you see, the function starts to have more branches and it's warning sign for maintainability. 

---
In reply to: [307847029](https://github.com/Azure/acr-builder/pull/469#discussion_r307847029) [](ancestors = 307847029)

builder/builder.go
```diff
@@ -236,14 +271,19 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step) error {
 			step.Build = replacePositionalContext(step.Build, ".")
 		}
 		step.UpdateBuildStepWithDefaults()
-		args = b.getDockerRunArgs(volName, workingDirectory, step, step.Envs, "", "docker build "+step.Build)
+
+		if step.UseBuildCacheForBuildStep() {
+			args = b.getDockerRunArgs(volName, workingDirectory, step, step.Envs, "", buildxImg+" build "+step.Build, step.ID)
+		} else {
+			args = b.getDockerRunArgs(volName, workingDirectory, step, step.Envs, "", dockerImg+" build "+step.Build, step.ID)
+		}
```
northtyphoon: add new testcase in builder_test to cover the case.

graph/step.go
```diff
@@ -105,6 +111,11 @@ func (s *Step) Validate() error {
 			return NewSelfReferencedStepError(fmt.Sprintf("Step ID: %v is self-referenced", s.ID))
 		}
 	}
+
+	if s.Cache != "" && strings.ToLower(s.Cache) != enabled && strings.ToLower(s.Cache) != disabled {
```
northtyphoon: >strings.ToLower(s.Cache) != enabled && strings.ToLower(s.Cache) != disabled [](start = 21, length = 75)

is it possible to use [https://golang.org/pkg/strings/#EqualFold](https://golang.org/pkg/strings/#EqualFold)?

graph/step.go
```diff
@@ -196,3 +208,55 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+	return fmt.Sprintf("cache_%s_%s", taskName, stepID)
+}
+
+// GetCmdWithCacheFlags adds buildx cache parameters to the cmd.
+func (s *Step) GetCmdWithCacheFlags(taskName string) (string, error) {
+	result := ""
+
+	if s == nil {
+		return result, errors.New("step is nil")
+	}
```
northtyphoon: is it possible?

graph/step.go
```diff
@@ -196,3 +208,55 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+
+	if len(s.Tags) == 0 {
+		return result, errors.New("atleast one tag is required to use build cache's cache registry")
```
northtyphoon: >atleast [](start = 29, length = 7)

`at least`?

graph/step.go
```diff
@@ -196,3 +208,55 @@ func (s *Step) UpdateBuildStepWithDefaults() {
...
+	}
+
+	if strings.ToLower(s.Cache) != enabled {
+		return result, errors.New("cache needs to be set to 'enabled' to use buildcache")
```
northtyphoon: >buildcache [](start = 71, length = 10)

`build cache`

cmd/acb/commands/exec/exec.go
```diff
@@ -112,6 +112,10 @@ var Command = cli.Command{
 			Name:  "set",
 			Usage: "set values on the command line (use --set multiple times or use commas: key1=val1,key2=val2)",
 		},
+		cli.StringFlag{
+			Name:  "task-name",
```
northtyphoon: >task-name [](start = 11, length = 9)

maybe just `--name`?

builder/context.go
shahzzzam: yep, i have created a new umbrella function to create run arguments for docker. That will also be internally used by creating run args for Step.

builder/builder.go
shahzzzam: @northtyphoon I have already covered the unit test for UseBuildCacheForBuildStep in `step_test.go`.

shahzzzam:
Hi @northtyphoon @shizhMSFT Can I get an ack on this PR?

It has been tested in dogfood

builder/builder.go
shahzzzam: ping

Issue #509, opened 2019-09-19
build/exec cmd: Add an "adhoc" run option


juliusl:
**What is the problem you're trying to solve**
Currently executing a container image through Tasks diverges from a user context in order to add extra features and some conveniences. This makes it difficult for users to reproduce on their own machine.

**Describe the solution you'd like**
For example in the tasks.yaml add an option as so 
```
- cmd: ..
  asAdhoc: true 
``` 

The runner should now execute the container without adding any of it's own context to the execution. As if the user was running the image adhoc on their own machine. 

**Additional Context**
This is useful if the user would like to execute their own scripts with a plain view of their container. Use case would be validation before the image makes it to their runtime.

juliusl:
So we discussed this and did some experimentation. We found a pretty solid way to achieve this and I wrote an example project here:

https://github.com/juliusl/pure-docker-run-on-tasks

PR #520, opened 2019-10-04
enhance debug mode for local debugging

shahzzzam:

**Purpose of the PR**

- 

Fixes #


sajayantony:
Let's see if we can just use docker run semantics for now.

juliusl:
I think we can drop this PR now, 

See:
#509

PR #521, opened 2019-10-04
avoid fetching digest for buildx

shahzzzam:

**Purpose of the PR**

- 

Fixes #

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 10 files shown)


graph/step.go
```diff
@@ -18,7 +18,7 @@ import (
 const (
 	// ImmediateExecutionToken defines the when dependency to indicate a step should execute immediately.
 	ImmediateExecutionToken = "-"
-	windowsOS               = "windows"
+	windowsOS               = util.WindowsOS
```
northtyphoon: the const seems redundant

builder/builder.go
```diff
@@ -149,7 +150,9 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 	var deps []*image.Dependencies
 	for _, step := range task.Steps {
 		log.Printf("Step ID: %v marked as %v (elapsed time in seconds: %f)\n", step.ID, step.StepStatus, step.EndTime.Sub(step.StartTime).Seconds())
-
+		if step.UseBuildCacheForBuildStep() && runtime.GOOS == util.LinuxOS {
```
northtyphoon: 1. Can you add some comments to explain the limitation?
2. Also why add an additional runtime os check? does it work for windows?

graph/step.go
shahzzzam: oops! good old copy and paste got me

northtyphoon:
Can you open an item for the dependency tracking issue when cache is enabled?

builder/builder.go
shahzzzam: no - but if its a Windows Task, and if the task file contains "cache: enabled"

then `step.UseBuildCacheForBuildStep()` will return true. However, since it is Windows - we do not error out. We fall back to normal build.

So in this scenario, if its windows - then we will check for dep anyway.

northtyphoon: I forget we fallback to docker build. Do we print out any warning to tell user `cache:enabled` doesn't work on windows?

shahzzzam: yes, https://github.com/Azure/acr-builder/blob/master/graph/task.go#L327

PR #531, opened 2019-10-08
update readme with debug instructions

shahzzzam:

**Purpose of the PR**

- This gives the ability to debug the Task runtime locally on your Linux machine.

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,3 +90,83 @@ $ acb render -f acb.yaml --values values.yaml
 ```
 
 If your template uses `.Run.ID` or other `.Run` variables, refer to the full list of parameters using `acb render --help`.
+
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,4 +89,58 @@ $ docker run -v $(pwd):/workspace --workdir /workspace -v /var/run/docker.sock:/
...
+                "--id",
+                "blah",
+                "--registry",
+                "samashah.azurecr.io"
+            ]
+        }
+    ]
+}
+```
+Note that if you have credentials to push something to the registry, it doesn't really work right now because we don't have a way to pass in "/home" volume.
```
northtyphoon: i cannot remember. Is it the statement still true?

PR #534, opened 2019-10-09
load images from cache if using buildx

shahzzzam:

**Purpose of the PR**
Adds the functionality to load images into docker store built with using buildx (cache: enabled flag).

If build contains no tags, we just append a buildx `--load` flag.

If build contains at least one tag, but no domain - we grab the Registry login server and create a cache image you can use in future for your builds.

If build contains at least one tag, and has a login server (fully qualified image name), we use that domain for image cache.

Fixes #

### DIFF
--- a/graph/step.go
+++ b/graph/step.go
@@ -233,39 +233,62 @@ func GetBuildCacheImageTag(taskName, stepID string) string {
 }
 
 // GetCmdWithCacheFlags adds buildx cache parameters to the cmd.
-func (s *Step) GetCmdWithCacheFlags(taskName string) (string, error) {
-	result := ""
-
-	if len(s.Tags) == 0 {
-		return result, errors.New("at least one tag is required to use build cache's cache registry")
-	}
+func (s *Step) GetCmdWithCacheFlags(taskName, registry string) (string, error) {
+	var domain, path, firstTagPath string
+	var err error
 
 	if strings.ToLower(s.Cache) != enabled {
-		return result, errors.New("cache needs to be set to 'enabled' to use build cache")
+		return "", errors.New("cache needs to be set to 'enabled' to use build cache")
+	}
+	if len(s.Tags) == 0 {
+		return s.Build, nil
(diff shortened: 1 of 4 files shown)


graph/step.go
```diff
@@ -257,8 +257,11 @@ func (s *Step) GetCmdWithCacheFlags(taskName string) (string, error) {
 	//  trim out any digests and tags
 	named = reference.TrimNamed(named)
 	domain, _ := reference.SplitHostname(named)
+	// todo: (sam) discuss if domain is empty, do we automatically attach $Registry?
+	// that might break behavior where customers use this image later on without $Registry
 	if domain == "" {
-		return result, errors.New("domain for the registry is required to push build cache")
+		result = fmt.Sprintf("--load %s", s.Build)
```
northtyphoon: if the build step has multiple tags like -t image1 -t image2, will the code load all of them?

graph/step.go
```diff
@@ -233,39 +233,62 @@ func GetBuildCacheImageTag(taskName, stepID string) string {
...
 // GetCmdWithCacheFlags adds buildx cache parameters to the cmd.
-func (s *Step) GetCmdWithCacheFlags(taskName string) (string, error) {
-	result := ""
+func (s *Step) GetCmdWithCacheFlags(taskName, registry string) (string, error) {
+	var foundDomain bool
+	var named reference.Named
 
+	if strings.ToLower(s.Cache) != enabled {
+		return "", errors.New("cache needs to be set to 'enabled' to use build cache")
+	}
```
northtyphoon: why do you need the check here? is it for debug purpose?

shahzzzam: yeah it should never really happen. But helps with unit tests

northtyphoon:
@shahzzzam next time, can you also fill the following instead of leaving it as blank?
```
Purpose of the PR
Fixes #
```
If the above format doesn't fit, remove it and add some description.

PR #545, opened 2019-11-21
Refactor load and render in templating

### LINKED ISSUE #540: acb build failed with tag name contains "secrets"
Run the following build command that uses a tag name contains "secrets"

```
acb build -t secretsimage:latest --push http://github/repo.git
```

it throws the error 

failed to resolve secrets in the task with error: failed to parse template to create task: yaml: unmarshal errors:
line 1: cannot unmarshal !!str `--pull ...` into graph.Task

jaysterp:
**Purpose of the PR**

- fixes an bug where `LoadAndRenderSteps` tries to render secrets from build parameters if they contain the string "secrets" (ie. the tag name), a functionality used for exec. This breaks during unmarshal as the build content is usually not valid yaml.
- refactor the load and render methods so that build does not try to resolve secrets. We will expand the load logic of exec in the future so we decided to separate the paths now.

Fixes #540

### DIFF
--- a/templating/base_render_options.go
+++ b/templating/base_render_options.go
@@ -127,7 +127,7 @@ func OverrideValuesWithBuildInfo(c1 *Config, c2 *Config, opts *BaseRenderOptions
 // and base render options.
 func LoadAndRenderBuildSteps(ctx context.Context, template *Template, opts *BaseRenderOptions) (string, error) {
 	// load steps and override values
-	mergedVals, err := LoadSteps(template, opts)
+	mergedVals, err := loadSteps(template, opts)
 	if err != nil {
 		return "", fmt.Errorf("error while loading build steps: %v", err)
 	}
@@ -150,7 +150,7 @@ func LoadAndRenderBuildSteps(ctx context.Context, template *Template, opts *Base
 // and base render options.
 func LoadAndRenderSteps(ctx context.Context, template *Template, opts *BaseRenderOptions) (string, error) {
 	// load steps and override values
-	mergedVals, err := LoadSteps(template, opts)
+	mergedVals, err := loadSteps(template, opts)
 	if err != nil {
 		return "", fmt.Errorf("error while loading exec steps: %v", err)
 	}
@@ -180,8 +180,8 @@ func LoadAndRenderSteps(ctx context.Context, template *Template, opts *BaseRende
 	return rendered, nil
 }
 
-// LoadSteps loads a template file and overrides values with build info
-func LoadSteps(template *Template, opts *BaseRenderOptions) (Values, error) {
+// loadSteps loads a template file and overrides values with build info
+func loadSteps(template *Template, opts *BaseRenderOptions) (Values, error) {
 	// return empty values list for an empty template.
 	if len(template.GetData()) == 0 {
 		return nil, nil
--- a/templating/base_render_options.go
+++ b/templating/base_render_options.go
@@ -224,7 +224,6 @@ func renderAndResolveSecrets(
 	opts *BaseRenderOptions,
 	sourceValues Values) (Values, error) {
 	result := Values{}
-
 	// Cheap optimization to skip the secrets merging if the yaml definition file doesn't contain "secrets" string in it. Note that the task can
 	// have the string secrets but may not essentially the secrets section.
 	if !strings.Contains(string(template.Data), "secrets") {
--- a/templating/base_render_options.go
+++ b/templating/base_render_options.go
@@ -154,6 +154,10 @@ func LoadAndRenderSteps(ctx context.Context, template *Template, opts *BaseRende
 	if err != nil {
 		return "", fmt.Errorf("error while loading exec steps: %v", err)
 	}
+	// return empty rendered string for an empty template.
+	if mergedVals == nil {
+		return "", nil
+	}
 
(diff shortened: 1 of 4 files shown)


templating/base_render_options.go
```diff
@@ -123,24 +123,80 @@ func OverrideValuesWithBuildInfo(c1 *Config, c2 *Config, opts *BaseRenderOptions
...
+
+	if rendered == "" {
+		return "", errors.New("rendered template was empty")
+	}
+
+	return rendered, nil
+}
+
+// LoadSteps loads a template file and overrides values with build info
+func LoadSteps(template *Template, opts *BaseRenderOptions) (Values, error) {
```
northtyphoon: >LoadSteps [](start = 5, length = 9)

if you don't plan to expose outside the module, lowercase loadSteps to make it private

templating/base_render_options.go
```diff
@@ -123,24 +123,80 @@ func OverrideValuesWithBuildInfo(c1 *Config, c2 *Config, opts *BaseRenderOptions
 	return base, nil
 }
 
-// LoadAndRenderSteps loads a template file and renders it according to an optional values file, --set values,
+// LoadAndRenderBuildSteps loads a template file for build and renders it according to an optional values file, --set values,
+// and base render options.
+func LoadAndRenderBuildSteps(ctx context.Context, template *Template, opts *BaseRenderOptions) (string, error) {
```
northtyphoon: Can you add unit tests for the new methods?

jaysterp: @northtyphoon Sure, thanks

templating/base_render_options.go
jaysterp: That's a cool go feature. Yes, this method should be private

templating/base_render_options_test.go
```diff
@@ -137,6 +137,40 @@ func TestLoadAndRenderSteps(t *testing.T) {
...
+func TestLoadAndRenderBuildSteps(t *testing.T) {
+	opts := &BaseRenderOptions{
+		ValuesFile: "",
+	}
+	tests := []struct {
+		buildFile string
+		expected  string
+	}{
+		{
+			"testdata/caching/Dockerfile-test",
```
northtyphoon: May I suggest use a real case for this testcase? In build command, the template string is something like

`--pull --label a=b --no-cache -f Dockerfile --bug-arg c=d -t {{.Run.Registry}} --platform linux/amd64 /workspace`

templating/base_render_options.go
shahzzzam: Can we just have a flag instead - called `renderSecrets bool`.

```suggestion
// LoadAndRenderSteps loads a template file and renders it according to an optional values file, --set values,
// and base render options.
func LoadAndRenderSteps(ctx context.Context, template *Template, opts *BaseRenderOptions, renderSecrets bool) (string, error) {
	// return empty rendered string for an empty template.
	if len(template.GetData()) == 0 {
		return "", nil
	}

	var err error

	config := &Config{}
	if opts.ValuesFile != "" {
		if config, err = LoadConfig(opts.ValuesFile); err != nil {
			return "", err
		}
	} else if opts.Base64EncodedValuesFile != "" {
		if config, err = DecodeConfig(opts.Base64EncodedValuesFile); err != nil {
			return "", err
		}
	}

	setConfig := &Config{}
	if len(opts.TemplateValues) > 0 {
		var rawVals string
		rawVals, err = parseValues(opts.TemplateValues)
		if err != nil {
			return "", err
		}

		setConfig = &Config{RawValue: rawVals, Values: map[string]*Value{}}
	}

	mergedVals, err := OverrideValuesWithBuildInfo(config, setConfig, opts)
	if err != nil {
		return "", fmt.Errorf("failed to override values: %v", err)
	}

	engine := NewEngine()
	
	if renderSecrets {
	    // we will pass nil for the secret resolve override so as to use the default resolve function.
	    secrets, err := renderAndResolveSecrets(ctx, template, engine, nil, opts, mergedVals)
	    if err != nil {
		    return "", fmt.Errorf("failed to resolve secrets in the task with error: %v", err)
	    }
	    // update the secrets collection with resolved secrets.
	    mergedVals["Secrets"] = secrets
	}
	
	rendered, err := engine.Render(template, mergedVals)
	if err != nil {
		return "", fmt.Errorf("error while rendering templates: %v", err)
	}

	if rendered == "" {
		return "", errors.New("rendered template was empty")
	}

	return rendered, nil
}
```

jaysterp: @shahzzzam as build and exec validations continue to evolve separately we decided to split the methods. /cc @northtyphoon

shahzzzam: Okay sounds good

templating/base_render_options_test.go
```diff
@@ -137,6 +137,40 @@ func TestLoadAndRenderSteps(t *testing.T) {
...
+			`FROM node:9-alpine
```
shahzzzam: nit:
can you start this from next line - just easier to read

templating/base_render_options_test.go
shahzzzam: yep, u can create a template using something like:

https://github.com/Azure/acr-builder/blob/master/cmd/acb/commands/build/build.go#L315

PR #546, opened 2019-11-22
handle quotes for docker build step args

jaysterp:
**Purpose of the PR**

- Allow the build step parser to handle quotes around arguments.

Fixes #537

PR #547, opened 2019-11-22
Handle quotes in Dockerfile FROM

### LINKED ISSUE #529: Tasks can't handle quotes in Dockerfile FROM cmd directive
**Describe the bug**

If at the top of your Dockerfile you do:

```
FROM "alpine"
```

Tasks will fail.

**To Reproduce**
1. Add quotes around your base image.

**Expected behavior**
Either the docker build succeeds or  it shows a better error message. Currently this will work with docker build. 

This is the error message:
```
2019/10/08 21:57:58 Output from dependency scanning: fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Failed to parse image reference, ensure tags have a valid format: "ubuntu":latest: invalid reference format
failed to run step ID: scratchscripts: failed to scan dependencies: exit status 1

Run ID: ycn failed after 6s. Error: failed during run, err: exit status 1
```

Note that this section are just warnings:
```
2019/10/08 21:57:58 Output from dependency scanning: fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

These messages should be cleaned up or clearly marked the type of log messages they represent.

**Screenshots**

**Any relevant environment information**
 - OS: [e.g. Windows]
 - Version [e.g. v1.0.0]

**Additional context**

jaysterp:
**Purpose of the PR**

- Handle quotes in Dockerfile FROM

Fixes #529

### DIFF
--- a/scan/dependencies.go
+++ b/scan/dependencies.go
@@ -127,7 +127,7 @@ func NormalizeImageTag(img string) string {
 
 // NewImageReference parses a path of a image and creates a ImageReference object
 func NewImageReference(imagePath string) (*image.Reference, error) {
-	ref, err := reference.Parse(imagePath)
+	ref, err := reference.Parse(util.TrimQuotes(imagePath))
 	if err != nil {
 		return nil, errors.Wrapf(err, "Failed to parse image reference, ensure tags have a valid format: %s", imagePath)
 	}
--- a/scan/dependencies.go
+++ b/scan/dependencies.go
@@ -127,7 +127,7 @@ func NormalizeImageTag(img string) string {
 
 // NewImageReference parses a path of a image and creates a ImageReference object
 func NewImageReference(imagePath string) (*image.Reference, error) {
-	ref, err := reference.Parse(util.TrimQuotes(imagePath))
+	ref, err := reference.Parse(imagePath)
 	if err != nil {
 		return nil, errors.Wrapf(err, "Failed to parse image reference, ensure tags have a valid format: %s", imagePath)
 	}
@@ -196,7 +196,8 @@ SCAN: // label for the scan loop
 				if len(tokens) < 2 {
 					return "", nil, fmt.Errorf("unable to understand line %s", line)
 				}
-				var imageToken = os.Expand(tokens[1], func(key string) string {
+				// trim surrounds single and double quotes from the image reference
+				var imageToken = os.Expand(util.TrimQuotes(tokens[1]), func(key string) string {
 					return context[key]
 				})
 				var found bool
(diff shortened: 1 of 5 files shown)


scan/dependencies.go
```diff
@@ -127,7 +127,7 @@ func NormalizeImageTag(img string) string {
 
 // NewImageReference parses a path of a image and creates a ImageReference object
 func NewImageReference(imagePath string) (*image.Reference, error) {
-	ref, err := reference.Parse(imagePath)
+	ref, err := reference.Parse(util.TrimQuotes(imagePath))
```
northtyphoon: I added a comment to TrimQuotes in https://github.com/Azure/acr-builder/pull/546#discussion_r349727992

shahzzzam: we should handle the quote here instead:

https://github.com/Azure/acr-builder/blob/master/scan/dependencies.go#L202

shahzzzam: Also, there is a function in this file that does the same, you might wanna move that to Util and use that instead:

https://github.com/Azure/acr-builder/blob/master/scan/dependencies.go#L283

Your function below will fail in this case:

```go
	func TrimQuotes(s string) string {
	if strings.HasPrefix(s, "'") {
		return strings.Trim(s, "'")
	}
	return strings.Trim(s, "\"")
}
```

FROM alpine' 

Since it checks for Prefix. We can just use this instead:

```go
// removeSurroundingQuotes trims double quotes, then single quotes.
func removeSurroundingQuotes(s string) string {
	s = strings.Trim(s, `"`)
	return strings.Trim(s, `'`)
}
```

jaysterp: @shahzzzam we can optimize by simply having
```
// removeSurroundingQuotes trims double quotes, then single quotes.
func removeSurroundingQuotes(s string) string {
	return strings.Trim(s, `'\"`)
}
```
See fix.

jaysterp: @northtyphoon @shahzzzam the `removeSurroundingQuotes` method breaks in the case where the beginning and ending quotes are of different type. ie. `"'hello world"'` renders as `hello world"`. I will move this method to util instead with the fix

jaysterp: @shahzzzam good suggestion. I will move the trim there when we first parse FROM

PR #555, opened 2020-05-08
Read 2048 magic bytes to accommodate extended pax

shahzzzam:
**Purpose of the PR**

- Read 2048 magic bytes to accommodate PAX tarfiles
https://github.com/docker/cli/pull/1770/files

Fixes #

### DIFF
--- a/scan/context.go
+++ b/scan/context.go
@@ -100,7 +100,8 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
 	buf := bufio.NewReader(r)
 	var magic []byte
 
-	magic, err = buf.Peek(archiveHeaderSize)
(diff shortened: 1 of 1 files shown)


scan/context.go
```diff
@@ -100,7 +100,8 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
 	buf := bufio.NewReader(r)
 	var magic []byte
 
-	magic, err = buf.Peek(archiveHeaderSize)
+	// note: (sam) read 2048 magic bytes to accomodate Extended Pax headers.
+	magic, err = buf.Peek(archiveHeaderSize * 4)
```
northtyphoon, line 104: dump question, if the tarball length is less than 2K, will Peek return EOF?

shahzzzam: Yes, Peek return "EOF" which is handled below.

PR #556, opened 2020-05-09
fix mis-prefix custom registry image tag in build

northtyphoon:
**Purpose of the PR**

In `acb build` command, the code always tries to prefix image tag with `registry`. However if the customer wants to build and push the image to another registry, the code unexpectedly alter the tag.

The change collects all known registries form the command option and skip the prefix if the tag already contains the registry.

### DIFF
--- a/graph/task.go
+++ b/graph/task.go
@@ -332,7 +332,7 @@ func (t *Task) initialize(ctx context.Context) error {
 				}
 			}
 		} else if s.IsPushStep() {
-			s.Push = getNormalizedDockerImageNames(s.Push, t.RegistryName)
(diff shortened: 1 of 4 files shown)


graph/task.go
```diff
@@ -361,7 +361,6 @@ func getNormalizedDockerImageNames(dockerImages []string, registry string) []str
 	normalizedDockerImages := []string{}
 	for _, dockerImage := range dockerImages {
 		d := scan.NormalizeImageTag(dockerImage)
-		d = util.PrefixRegistryToImageName(registry, d)
```
northtyphoon, line 364: The code never takes any effect.

PR #558, opened 2020-05-22
fix build issue with linter upgrade

### LINKED ISSUE #533: Upgrade golangci-lint to 1.20
golangci-lint 1.20 reported the following error and failed the build

pkg/procmanager/procmanager.go:111: G204: Subprocess launched with function call as argument or cmd arguments (gosec)
        cmd := exec.Command(args[0], args[1:]...)
Makefile:71: recipe for target 'lint' failed

akashsinghal:
**Purpose of the PR**

- added linter tag so gosec ignores build rule G204 just for line of code causing build to fail
- upgraded golangci-lint to v1.20.1
- working on fixing errors produced by upgrading to v1.27.0

Fixes #533

### DIFF
--- a/cmd/acb/commands/exec/exec.go
+++ b/cmd/acb/commands/exec/exec.go
@@ -150,7 +150,7 @@ var Command = cli.Command{
 			taskFile = defaultTaskFile
 		}
 
-		ctx := gocontext.WithValue(gocontext.Background(), "debug", debug)
+		ctx := gocontext.Background()
 		pm := procmanager.NewProcManager(dryRun)
 
 		if homevol == "" {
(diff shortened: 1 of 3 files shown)


shahzzzam:
LGTM for 1.20, can you post the errors from version v1.27.0?

akashsinghal:
Here's the error message from staticcheck when upgrading to 1.27.0:

cmd/acb/commands/exec/exec.go:153:54: SA1029: should not use built-in type string as key for value; define your own type to avoid collisions (staticcheck)
		ctx := gocontext.WithValue(gocontext.Background(), "debug", debug)

shahzzzam:
Thanks, I see. 

Do you think we can remove putting the "debug" flag in the `ctx` altogether? it seems unnecessary and the bool `debug` can do the same job.

https://github.com/Azure/acr-builder/blob/cdd23eeaae850ac05df17588857589bf4b799bd7/cmd/acb/commands/exec/exec.go#L131

PR #560, opened 2020-06-22
Mount Secrets As File

### LINKED ISSUE #550: Support "mounting" values and secrets through file
**What is the problem you're trying to solve**

Allow passing values and secrets to each step through file

```
volumes:
- name: foo
  values: 
   mysecret1: {{.Secrets.mysecret1}}  

steps:
- cmd: image
  mounts: 
  - name: foo
    mountPath: '/run/foo'
```

akashsinghal:
**Purpose of the PR**

- New feature to allow users to specify volumes to mount into containers at each step definition

Fixes #550 https://github.com/Azure/acr/issues/348


.golangci.yml
```diff
@@ -28,4 +28,7 @@ issues:
   exclude-rules:
     - text: "Potential HTTP request made with variable url"
       linters:
-        - gosec
\ No newline at end of file
+        - gosec
+    - text: "type name will be used as volume.VolumeMount by other packages, and that stutters; consider calling this Mount"
```
shahzzzam, line 32: Let's remove this. I have a suggestion.

The original `Volume.go` does not really describe a docker volume as mentioned in `pkg/volume/volume.go`. 
 
I suggest renaming the file/struct as `DockerVolumeHelper` or something. Since it is nothing but a wrapper on calling helper methods to create/remove docker volume.

And once you do that. You can rename `volume_mount.go` to `volume.go` and `VolumeMount` to `Volume` which is more appropriate IMO.

What do you think?

akashsinghal: Sounds good to me. I'll refactor it.

shahzzzam:
Another suggestion for `volume.go` design.

I would suggesting defining a struct like:

```go
pkg/volume/source.go
----
package volume

// Represents the source of a volume to mount.
// Only one of its members may be specified.
type Source struct {
	// Secret represents a secret that should populate this volume.
	Secret []map[string]string `yaml:"secret"`
   
        // add more sources here ...
}
```

This way, later on: we can add more sources as we want. And it is easy to extend. 

K8s supports a lot of different types of sources: Check out their [source code](https://github.com/kubernetes/kubernetes/blob/9859793a8b6b8961d36cefb6feaed4209b01f0af/staging/src/k8s.io/api/core/v1/types.go#L38) (one of them is SecretVolumeSource).

And then you can refer to the source like this:

```go
type Volume struct {
	Name   string               `yaml:"name"`
	// Source represents the location and type of the mounted volume.
	Source Source `yaml:",inline"`
}
```

What do you think?

pkg/volume/volume.go
```diff
@@ -4,49 +4,40 @@
...
-	VolumePrefix = "acb_vol_"
-)
-
-// Volume describes a Docker volume.
+// Volume describes a Docker bind mounted volume.
 type Volume struct {
-	Name        string
-	procManager *procmanager.ProcManager
+	Name   string              `yaml:"name"`
+	Values []map[string]string `yaml:"secret"`
```
shahzzzam, line 15: Source?

builder/builder.go
```diff
@@ -504,14 +504,14 @@ func (b *Builder) createFilesForVolume(ctx context.Context, volMount *volume.Vol
...
 
-func (b *Builder) populateVolumeWithFiles(ctx context.Context, volMount *volume.VolumeMount) error {
+func (b *Builder) populateVolumeWithFiles(ctx context.Context, volMount *volume.Volume) error {
 	var dataContainerArgs []string
 	var dataSB strings.Builder
 	if runtime.GOOS == util.WindowsOS {
 		dataContainerArgs = []string{"powershell.exe", "-Command"}
-		dataSB.WriteString("docker run --rm -v " + b.workspaceDir + "\\" + volMount.Name + ":c:\\source -v ")
+		dataSB.WriteString("docker run --rm -v " + b.workspaceDir + ":c:\\source -v ")
 		dataSB.WriteString(volMount.Name + ":c:\\dest -w /source ")
```
shahzzzam, line 513: -w c:\\source
?

builder/builder.go
```diff
@@ -504,14 +504,14 @@ func (b *Builder) createFilesForVolume(ctx context.Context, volMount *volume.Vol
...
 }
 
-func (b *Builder) populateVolumeWithFiles(ctx context.Context, volMount *volume.VolumeMount) error {
+func (b *Builder) populateVolumeWithFiles(ctx context.Context, volMount *volume.Volume) error {
 	var dataContainerArgs []string
 	var dataSB strings.Builder
 	if runtime.GOOS == util.WindowsOS {
 		dataContainerArgs = []string{"powershell.exe", "-Command"}
-		dataSB.WriteString("docker run --rm -v " + b.workspaceDir + "\\" + volMount.Name + ":c:\\source -v ")
+		dataSB.WriteString("docker run --rm -v " + b.workspaceDir + ":c:\\source -v ")
```
shahzzzam, line 512: -w c:\source
?

akashsinghal: It seems to work without that as well locally. Do you think I should change it anyways?

builder/builder.go
```diff
@@ -504,14 +504,14 @@ func (b *Builder) createFilesForVolume(ctx context.Context, volMount *volume.Vol
 	return nil
 }
 
-func (b *Builder) populateVolumeWithFiles(ctx context.Context, volMount *volume.VolumeMount) error {
+func (b *Builder) populateVolumeWithFiles(ctx context.Context, volMount *volume.Volume) error {
```
shahzzzam, line 507: Giving this more thought, why do we need to `populateVolumeWithFiles`. Could you remind me?

We are using bind mounts **(not docker volume mount)** anyway. I think we can drop this entire method?

Edit:

When you start creating files for Volume.

createFilesForVolume(), lets create all the files in a folder called `volumeSource`.

So, `mkdir volumeSources` first, and then mkdir your volMount.Name.

This will have a cleaner folder structure. All the volumes will be inside the folder `volumeSources`. Now, 

```
volumes:
  - name: mysecrets
    secret:
      - mysecret1: {{"this is a non encoded string1" | b64enc}}
      - mysecret2: {{"this is a non encoded string2" | b64enc}}
```

After calling `createFilesForVolume`, folder structure on your host machine is:

```
my_acb_home_vol (b.workspaceDir)
|__volumeSources
       |__mysecrets
            |__mysecret1
            |__mysecret2
```

In `context.go` when you get docker run args:

```
if len(volMounts) > 0 {
		for key, val := range volMounts {
			sb.WriteString(" --volume " + "volumeSources/" + key + ":" + val)
		}
	}
```

Just an idea, can this work?

builder/builder.go
```diff
@@ -443,3 +459,77 @@ func parseImageNameFromArgs(cmdArgs string) string {
 	}
 	return cmdArgs[:idx]
 }
+
+func (b *Builder) createFilesForVolume(ctx context.Context, volMount *volume.Volume) error {
```
shahzzzam, line 463: name this differently:

`processVolumeSource()`

Ideally, it should process the source according to source type:
```go
case source.Secret != nil:
            // base64 decode
            // dump content on to a file
case source.AzureFile != nil:   (future scenario)
           // download source code from remote
           // create a whole new file system
default:
    throw err("not supported yet") 
```

In this way, this method is more extensible and can process different volume source types differently.

builder/context.go
```diff
@@ -87,6 +88,11 @@ func (b *Builder) getDockerRunArgs(
 	sb.WriteString(" --volume " + volName + ":" + containerWorkspaceDir)
 	sb.WriteString(" --volume " + util.DockerSocketVolumeMapping)
 	sb.WriteString(" --volume " + homeVol + ":" + homeWorkDir)
+	if len(volMounts) > 0 {
+		for key, val := range volMounts {
+			sb.WriteString(" --volume " + key + ":" + val)
```
shahzzzam, line 93: lets also mount this as readonly

append `:ro` at the end of it.

https://docs.docker.com/storage/bind-mounts/#use-a-read-only-bind-mount

This will avoid any containers to update the file content.

pkg/volume/volume.go
```diff
@@ -4,49 +4,31 @@
...
-func (v *Volume) getDockerCreateArgs() []string {
-	return []string{"docker", "volume", "create", "--name", v.Name}
-}
-
-func (v *Volume) getDockerRmArgs() []string {
-	return []string{"docker", "volume", "rm", v.Name}
+	if v.Name == "" {
+		return errors.New("volume name is empty")
+	}
+	if v.Source.Secret == nil {
```
shahzzzam, line 26: can you call v.Source.Validate() ? I see you have a Validate method in source.go

pkg/volume/source.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/pkg/errors"
+)
+
+// Source represents the source of a volume to mount.
+// Only one of its members may be specified.
+type Source struct {
+	// Secret represents a secret that should populate this volume.
+	Secret []map[string]string `yaml:"secret,omitempty"`
+
+	// add more sources here ...
```
shahzzzam, line 16: can remove for extra lines/comment for now

pkg/volume/volume.go
```diff
@@ -4,49 +4,31 @@
...
+	Name   string `yaml:"name"`
+	Source Source `yaml:",inline"`
 }
 
-// NewVolume creates a new Volume.
-func NewVolume(name string, pm *procmanager.ProcManager) *Volume {
-	return &Volume{
-		Name:        name,
-		procManager: pm,
+//Validate checks whether Volume is well formed
```
shahzzzam, line 18: nit: some of the code have a space after `//` and some of them doesn't...
can you add a space after `//` everywhere throughout the pr

builder/builder.go
shahzzzam: .. or `prepareVolumeSource()`

builder/builder.go
shahzzzam: talked offline. We need named volumes

graph/step.go
```diff
@@ -134,6 +146,39 @@ func (s *Step) Validate() error {
 	return nil
 }
 
+//ValidateMounts checks each mount is well formed and each container file path is unique
```
akashsinghal, line 149: Add initial space here for comment

graph/step.go
```diff
@@ -134,6 +146,39 @@ func (s *Step) Validate() error {
...
+		if _, exists := duplicate[m.MountPath]; exists {
+			return errors.New("mount with duplicate container file path found")
+		}
+
+		duplicate[m.MountPath] = struct{}{}
+	}
+	return nil
+}
+
+//ValidateMountVolumeNames checks mount name matches a listed volume
```
akashsinghal, line 167: Add a space here

graph/step.go
```diff
@@ -134,6 +146,39 @@ func (s *Step) Validate() error {
...
+func (s *Step) ValidateMountVolumeNames(vols []*volume.Volume) error {
+	//for each mount in the step, check to see that there exists a matching
```
akashsinghal, line 169: add a space here

graph/task.go
```diff
@@ -199,6 +201,17 @@ func (t *Task) Validate() error {
 
 		idMap[secret.ID] = struct{}{}
 	}
+
+	//Validate Volumes if exists
```
akashsinghal, line 205: Fix the spacing for comments in this file!

templating/engine.go
```diff
@@ -28,9 +30,20 @@ func NewEngine() *Engine {
...
+	if v == nil {
+		return ""
+	}
+	return base64.StdEncoding.EncodeToString([]byte(reflect.ValueOf(v).String()))
+}
+
 // FuncMap returns a FuncMap representing all of the functionality of the engine.
 func FuncMap() template.FuncMap {
-	return sprig.TxtFuncMap()
+	//We are overriding the b64enc function with custom implementation
```
akashsinghal, line 43: Add a space in the comment here

builder/builder.go
```diff
@@ -443,3 +454,92 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+			}
+		}
+		dataSB.WriteString("/dest")
+	}
+	dataContainerArgs = append(dataContainerArgs, dataSB.String())
+	var buf bytes.Buffer
+	if err := b.procManager.Run(ctx, dataContainerArgs, nil, &buf, &buf, ""); err != nil {
+		return errors.Wrapf(err, "failed to populate container, %s", buf.String())
+	}
+	log.Println("buffer: ", buf.String())
```
akashsinghal, line 542: Will be removed before submitting for prod

builder/builder.go
```diff
@@ -443,3 +454,92 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+				sb.WriteString("\nEOL\n")
+			}
+		}
+	}
+	args = append(args, sb.String())
+	var buf bytes.Buffer
+	if err := b.procManager.Run(ctx, args, nil, &buf, &buf, ""); err != nil {
+		return errors.Wrapf(err, "failed to write value, %s", buf.String())
+	}
+	log.Println("buffer: ", buf.String())
```
akashsinghal, line 511: Will be removed before submitting for prod

builder/builder.go
```diff
@@ -443,3 +454,92 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+	var dataSB strings.Builder
+	if runtime.GOOS == util.WindowsOS {
+		dataContainerArgs = []string{"powershell.exe", "-Command"}
+		dataSB.WriteString("docker run --rm -v " + b.workspaceDir + ":c:\\source -v ")
+		dataSB.WriteString(volMount.Name + ":c:\\dest -w c:\\source ")
+		dataSB.WriteString(nanoServerImageName + " cmd.exe /c copy c:\\source\\" + volMount.Name + " c:\\dest")
+	} else {
+		dataContainerArgs = []string{"/bin/sh", "-c"}
+		dataSB.WriteString("docker run --rm -v " + b.workspaceDir + ":/source -v ")
+		dataSB.WriteString(volMount.Name + ":/dest -w /source alpine cp ")
```
northtyphoon, line 528: >alpine [](start = 56, length = 6)

is it possible to use bash? 

[https://github.com/Azure/acr-builder/blob/master/builder/constants_linux.go](https://github.com/Azure/acr-builder/blob/master/builder/constants_linux.go)

builder/builder.go
```diff
@@ -443,3 +454,92 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+
+// populateSecretVolume mounts all files of type source Secret generated into a volume
+func (b *Builder) populateSecretVolume(ctx context.Context, volMount *volume.Volume) error {
+	var dataContainerArgs []string
+	var dataSB strings.Builder
+	if runtime.GOOS == util.WindowsOS {
+		dataContainerArgs = []string{"powershell.exe", "-Command"}
+		dataSB.WriteString("docker run --rm -v " + b.workspaceDir + ":c:\\source -v ")
+		dataSB.WriteString(volMount.Name + ":c:\\dest -w c:\\source ")
+		dataSB.WriteString(nanoServerImageName + " cmd.exe /c copy c:\\source\\" + volMount.Name + " c:\\dest")
```
northtyphoon, line 524: >nanoServerImageName [](start = 21, length = 19)

Do you think we can consolidate the windows configure image to "mcr.microsoft.com/windows/nanoserver:2004"?

[https://github.com/Azure/acr-builder/blob/master/builder/constants_windows.go#L13](https://github.com/Azure/acr-builder/blob/master/builder/constants_windows.go#L13)

graph/step.go
```diff
@@ -134,6 +146,39 @@ func (s *Step) Validate() error {
...
+	duplicate := make(map[string]struct{}, len(mounts))
+	for _, m := range mounts {
+		// call m.Validate() for each mount
+		if err := m.Validate(); err != nil {
+			return err
+		}
+		// make sure each container file path provided is unique
+		if _, exists := duplicate[m.MountPath]; exists {
+			return errors.New("mount with duplicate container file path found")
+		}
```
northtyphoon, line 160: is it ok to have two mount path like the following?

/home/user
/home/user/john

graph/step.go
```diff
@@ -134,6 +146,39 @@ func (s *Step) Validate() error {
...
+// ValidateMountVolumeNames checks mount name matches a listed volume
+func (s *Step) ValidateMountVolumeNames(vols []*volume.Volume) error {
+	// for each mount in the step, check to see that there is a matching volume
+	nameMap := make(map[string]struct{}, len(vols))
+	for _, v := range vols {
+		nameMap[v.Name] = struct{}{}
+	}
+	for _, m := range s.Mounts {
+		if _, exists := nameMap[m.Name]; !exists {
+			return errors.New("provided mount name does not correspond to a volume")
```
northtyphoon, line 176: >provided mount name does not correspond to a volume [](start = 22, length = 51)

can you include the mount name in the error message?

builder/builder.go
akashsinghal: I can try replacing the config Image name from servercore to nanoserver and see if the config still works.

pkg/volume/source.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"github.com/pkg/errors"
+)
+
+// Source represents the source of a volume to mount.
+// Only one of its members may be specified.
+type Source struct {
+	// Secret represents a secret that should populate this volume.
+	Secret []map[string]string `yaml:"secret,omitempty"`
```
northtyphoon, line 14: >[]map[string]string [](start = 8, length = 19)

When I see the array of map, I notice the secret yaml is defined as a list of key:value pair.

  - name: mysecrets
    secret:
      - mysecret1: {{.Secrets.sampleSecret | b64enc}}
      - mysecret2: {{"this is a non encoded string" | b64enc}}

 Why we choose it rather than the dictionary like 

  - name: mysecrets
    secret:
      mysecret1: {{.Secrets.sampleSecret | b64enc}}
      mysecret2: {{"this is a non encoded string" | b64enc}}

graph/step.go
akashsinghal: Docker daemon allows for this since it considers it 2 separate directories. This is the same case as if the customer tried to mount the volume to the /workspace folder. You can't mount to /workspace directly since that's the home volume. But you can mount to a subdirectory like /workspace/sub

pkg/volume/source.go
akashsinghal: Yes I originally tried just using a map[string]string but that wasn't working for some reason so I used an array of maps instead as a workaround. I'll try the single map way again and try to make it work.

graph/step.go
akashsinghal: Upon further investigation, this scenario causes issues when we mount volumes as readonly

builder/builder.go
```diff
@@ -443,3 +453,88 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+			sb.WriteString("\nEOL\n")
+		}
+	}
+	args = append(args, sb.String())
+	var buf bytes.Buffer
+	if err := b.procManager.Run(ctx, args, nil, &buf, &buf, ""); err != nil {
+		return errors.Wrapf(err, "failed to write value, %s", buf.String())
+	}
+	log.Println("buffer: ", buf.String())
+	log.Println("Created new file(s) for volume: ", volMount.Name)
```
shahzzzam, line 509: @northtyphoon do we want to log this? or this sounds like an implementation detail?

builder/builder.go
```diff
@@ -443,3 +453,88 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+}
+
+// createSecretFiles creates necessary files for source type Secret
+func (b *Builder) createSecretFiles(ctx context.Context, volMount *volume.Volume) error {
+	var args []string
+	var sb strings.Builder
+	if runtime.GOOS == util.WindowsOS {
+		args = []string{"powershell.exe", "-Command"}
+		sb.WriteString("mkdir " + volMount.Name)
+		log.Println("making directory " + volMount.Name)
```
shahzzzam, line 477: do we need this log?

builder/builder.go
```diff
@@ -443,3 +453,88 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+		}
+		dataSB.WriteString("/dest")
+	}
+	dataContainerArgs = append(dataContainerArgs, dataSB.String())
+	var buf bytes.Buffer
+	if err := b.procManager.Run(ctx, dataContainerArgs, nil, &buf, &buf, ""); err != nil {
+		return errors.Wrapf(err, "failed to populate container, %s", buf.String())
+	}
+	log.Println("buffer: ", buf.String())
+	log.Println("Populated files for volume: ", volMount.Name)
```
shahzzzam, line 538: I guess instead of "populated" can we use something more generic like:

"Volume source `volMount.Name` successfully attached/mounted"

builder/builder.go
```diff
@@ -130,6 +133,13 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 		}
 	}
 
+	for _, volMount := range task.Volumes {
+		// create and populate volume for specified source
+		if err := b.prepareVolumeSource(ctx, volMount); err != nil {
+			return err
+		}
```
shahzzzam, line 140: How about just 1 log here

```
for _, volMount := range task.Volumes {
		// create and populate volume for specified source
		if err := b.prepareVolumeSource(ctx, volMount); err != nil {
			return errors.WithMessagef(err, "failed to created volume %q", volMount.Name)
		} 
		log.Printf("Volume source %q successfully created", volMount.Name)
	}
```

builder/builder.go
```diff
@@ -443,3 +453,88 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+func (b *Builder) createSecretFiles(ctx context.Context, volMount *volume.Volume) error {
+	var args []string
+	var sb strings.Builder
+	if runtime.GOOS == util.WindowsOS {
+		args = []string{"powershell.exe", "-Command"}
+		sb.WriteString("mkdir " + volMount.Name)
+		log.Println("making directory " + volMount.Name)
+	} else {
+		args = []string{"/bin/sh", "-c"}
+		sb.WriteString("mkdir " + volMount.Name + " && ")
```
shahzzzam, line 480: move the `&&` to below before `cat`. To be more consistent with windows `;` for better readability

I also recommend if it is possible to move the initializer to a file

like
init_shell_linux.go

args := []string{"/bin/sh", "-c"}

and init_shell_windows.go for powershell.

The `mkdir` line will be common for both of them

builder/builder.go
akashsinghal: i will remove it

PR #562, opened 2020-07-16
Add Artifact Support for Build Context

akashsinghal:
**Purpose of the PR**

- Add artifacts stored on registry as a third type of remote source/build context


util/trim.go
```diff
@@ -5,8 +5,14 @@ package util
...
+// TrimQuotes returns a slice of the string s, with all leading
 // and trailing double or single quotes removed, as defined by Unicode.
 func TrimQuotes(s string) string {
 	return strings.Trim(s, "'\"")
 }
+
+// TrimArtifactPrefix returns a slice of the string s, with leading
+// artifact prefix removed
+func TrimArtifactPrefix(s string) string {
+	return strings.TrimPrefix(s, "oci://")
```
northtyphoon, line 17: >oci:// [](start = 31, length = 6)

Is the schema like `OCI://`, "HTTPS://" valid? If yes, can you handle it in a case insensitive way?

builder/context.go
```diff
@@ -234,6 +237,15 @@ func getScanArgs(
 		args = append(args, "--build-arg", buildArg)
 	}
 
+	for _, cred := range credentials {
+		if serializedCredential, err := cred.String(); err != nil {
+			// Log error here, but throw it from scanner
+			log.Panicf("Credential validation failed for Source registry credential, err: %+v", err)
```
northtyphoon, line 243: >Source [](start = 48, length = 6)

nit: source

northtyphoon, line 243: >Panicf [](start = 7, length = 6)

Do we use panic in other places?

northtyphoon:
> 	}

I want to make sure we don't leak the possible sensitive info in our log, can you double check?

---
Refers to: builder/context.go:189 in 92f77b6. [](commit_id = 92f77b619f94bbc5329ac21f19897602705d5567, deletion_comment = False)

go.mod
```diff
@@ -1,8 +1,11 @@
 module github.com/Azure/acr-builder
 
+replace (
+	github.com/docker/docker => github.com/docker/docker v0.0.0-20180413235638-61138fb5fc0f
+)
 require (
 	github.com/Azure/azure-sdk-for-go v31.2.0+incompatible
-	github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78 // indirect
+	github.com/Azure/go-autorest v12.2.0+incompatible
```
northtyphoon, line 8: >github [](start = 1, length = 6)

Do you need to update vendor folder?

cmd/acb/commands/download/download.go
```diff
@@ -40,24 +41,41 @@ var Command = cli.Command{
...
+			return err
+		}
+
 		ctx, cancel := gocontext.WithTimeout(gocontext.Background(), timeout)
 		defer cancel()
 
 		pm := procmanager.NewProcManager(dryRun)
-		scanner, err := scan.NewScanner(pm, downloadCtx, "", destination, nil, nil, "")
+
+		registryLoginCredentials, err := graph.ResolveCustomRegistryCredentials(ctx, credentials)
```
northtyphoon, line 73: >ResolveCustomRegistryCredentials [](start = 41, length = 32)

Is there any perf penalty here if the source is not registry artifact? ResolveCustomRegistryCredentials may call msi and kv to get the credential? @shahzzzam

builder/context.go
akashsinghal: So I took this from how the runner in [worker.go](https://msazure.visualstudio.com/AzureContainerRegistry/_git/Build?path=%2Fsrc%2Facr-build-runner%2Fscheduler%2Fworker.go&version=GBakashsinghal%2FbuilderArtifactSupport&line=534&lineEnd=535&lineStartColumn=1&lineEndColumn=1&lineStyle=plain) handles credential serialization. Here they log the error rather than throw it since the function it is called in doesn't return an  error. If you think its better to throw an error, I can change the `getScanArgs` method to return an error if it there is one?

akashsinghal: Yea I had taken this from another piece of code. I will make some changes to the code so it return an error instead.

cmd/acb/commands/download/download.go
akashsinghal: I assume this is regarding the case where there are credentials passed in but we don't actually need those credentials because the context to download from is git or tar. In that case what I could add is to check here if the `downloadCtx` is actually a registry artifact and if it is then we can do all the credential processing.

shahzzzam: @northtyphoon yes, you;'re right. It will call Token server. @akashsinghal , correct. You can check for OCI, and only then pass in credentials to Scanner.

scan/context.go
```diff
@@ -113,6 +120,29 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
 	return err
 }
 
+func (s *Scanner) getContextFromRegistry(ctx context.Context, registryArtifact string) (err error) {
+	resolver := docker.NewResolver(docker.ResolverOptions{
+		PlainHTTP: false,
+		Client:    http.DefaultClient,
+		Credentials: func(hostName string) (string, string, error) {
```
shahzzzam, line 127: Suggestion: docker.Credentials is deprecated. Can you try using RegistryHosts..

I am not sure if it works, but we can try
```go
authorizer  := docker.NewAuthorizer(http.DefaultClient, func(hostName string) (string, string, error) {
    // If no matching credential found, attempt an anonymous pull
	if s.credentials[hostName] == nil {
		return "", "", nil
	}
	return s.credentials[hostName].Username.ResolvedValue, s.credentials[hostName].Password.ResolvedValue, nil
})
resolver := docker.NewResolver(docker.ResolverOptions{
        PlainHTTP: false,
	Hosts: docker.ConfigureDefaultRegistries(
		docker.WithAuthorizer(authorizer),
	),
})
```

akashsinghal: Switched to this and works locally.

PR #565, opened 2020-08-12
Fix for Multi File Volume

akashsinghal:
**Purpose of the PR**

- Fix issue where volume defined with more than one file in Linux containers would fail

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -492,7 +492,7 @@ func (b *Builder) createSecretFiles(ctx context.Context, volMount *volume.Volume
 		if runtime.GOOS == util.WindowsOS {
 			sb.WriteString("Add-Content -Path ")
 			sb.WriteString(volMount.Name + "/" + k)
-			sb.WriteString(" -Value @\"\n")
+			sb.WriteString(" -Value @\"\r\n")
 			sb.WriteString(val)
 			sb.WriteString("\n\"@")
 		} else {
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -474,34 +474,39 @@ func (b *Builder) prepareVolumeSource(ctx context.Context, volMount *volume.Volu
...
 		decoded, err := base64.StdEncoding.DecodeString(val)
 		if err != nil {
 			return errors.New("failed to decode Base64 value. please make sure value provided is Base64 encoded")
 		}
 		val = string(decoded)
 		if runtime.GOOS == util.WindowsOS {
-			sb.WriteString("; Add-Content -Path ")
+			sb.WriteString("Add-Content -Path ")
 			sb.WriteString(volMount.Name + "/" + k)
 			sb.WriteString(" -Value @\"\n")
```
northtyphoon, line 495: windows newline is usually `\r\n`

akashsinghal: switched to \r\n

builder/builder.go
```diff
@@ -474,34 +474,39 @@ func (b *Builder) prepareVolumeSource(ctx context.Context, volMount *volume.Volu
 // createSecretFiles creates necessary files for source type Secret
 func (b *Builder) createSecretFiles(ctx context.Context, volMount *volume.Volume) error {
 	var args []string
-	var sb strings.Builder
 	args = getShell()
-	sb.WriteString("mkdir " + volMount.Name)
+	args = append(args, "mkdir "+volMount.Name)
+	var buf bytes.Buffer
+	if err := b.procManager.Run(ctx, args, nil, &buf, &buf, ""); err != nil {
+		return errors.Wrapf(err, "failed to write value, %s", buf.String())
```
shahzzzam, line 481: update the error message

PR #570, opened 2020-11-11
fix acr build for git links without master branch

xyxyxyxyxyxy:
**Purpose of the PR**

The default branch of a git repo is not always "master".
github.com/repo.git => we should always use default branch no matter master exists or not or master is default or not
github.com/repo.git#dev => we should respect the settings, if dev doesn't exist, the download code should return error.
- 

Fixes #

### DIFF
--- a/scan/git.go
+++ b/scan/git.go
@@ -151,7 +151,8 @@ func checkoutGit(root, ref, subdir string) (string, error) {
 	// Try checking out by ref name first. This will work on branches and sets
 	// .git/HEAD to the current branch name
 	if output, err := gitWithinDir(root, "checkout", ref); err != nil {
-		// If the branch name is specified, throw an error
+		// If the branch name is specified, then it means the branch does not exist,
+		// so throw an error
 		if ref != "" {
 			return "", errors.Wrapf(err, "error checking out %s: %s", ref, output)
 		}
--- a/scan/git.go
+++ b/scan/git.go
@@ -151,9 +151,14 @@ func checkoutGit(root, ref, subdir string) (string, error) {
 	// Try checking out by ref name first. This will work on branches and sets
 	// .git/HEAD to the current branch name
 	if output, err := gitWithinDir(root, "checkout", ref); err != nil {
-		// If checking out by branch name fails check out the last fetched ref
-		if _, err2 := gitWithinDir(root, "checkout", "FETCH_HEAD"); err2 != nil {
+		// If the branch name is specified, throw an error
+		if ref != "" {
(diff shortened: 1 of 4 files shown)


juliusl (approved):

scan/git.go
```diff
@@ -151,10 +151,15 @@ func checkoutGit(root, ref, subdir string) (string, error) {
 	// Try checking out by ref name first. This will work on branches and sets
 	// .git/HEAD to the current branch name
 	if output, err := gitWithinDir(root, "checkout", ref); err != nil {
-		// If checking out by branch name fails check out the last fetched ref
-		if _, err2 := gitWithinDir(root, "checkout", "FETCH_HEAD"); err2 != nil {
+		// If the branch name is specified, throw an error
```
northtyphoon, line 154: @xyxyxyxyxyxy I know we never unit test the class before. But I'm wondering if it's possible to unit test it? /cc: @juliusl @shahzzzam

xyxyxyxyxyxy: This file refers to https://github.com/moby/moby/blob/master/builder/remotecontext/git/gitutils.go a lot.
The moby project has a corresponding test file https://github.com/moby/moby/blob/master/builder/remotecontext/git/gitutils_test.go
Maybe we can refer to that.

huanwu, line 154: Why not allow specify the branch name?

xyxyxyxyxyxy: It means we try to checkout to that branch but fail. Therefore, we throw an error.

xyxyxyxyxyxy: Unit tests added.

scan/git_test.go
```diff
@@ -0,0 +1,330 @@
...
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"gotest.tools/v3/assert"
+	is "gotest.tools/v3/assert/cmp"
+)
+
+func TestParseRemoteURL(t *testing.T) {
```
northtyphoon, line 24: Thanks for the unit test. Can you also add some testcases for devops? You can some url example from url_test.go

https://github.com/Azure/acr-builder/blob/627fa8df6a16b8742f7a55daf90415a6dce100e6/util/url_test.go#L22

northtyphoon (changes requested):

scan/git_test.go
xyxyxyxyxyxy: Added

northtyphoon (approved):

Issue #574, opened 2021-04-01
Scanning for dependencies fail when building using docker BUILDKIT


mnltejaswini:
**Describe the bug**
When building with DOCKER_BUILDKIT=1, the build fails to scan dependencies with the following error 
```
failed to query digests, msg:
Error: No such object: mcr.microsoft.com/dotnet/sdk:3.1
: exit status 1
```

**To Reproduce**
1. create a simple Dockerfile 
Sample Dockerfile
```Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:3.1
```

2. create a task file to build that docker file with BUILDKIT enabled 
```yaml
version: v1.1.0
steps:
  - id: build
    build: -t $Registry/test:$ID -f Dockerfile .
    env: ["DOCKER_BUILDKIT=1"]
```

3. The build fails to scan dependencies.
```failed to query digests, msg:
Error: No such object: mcr.microsoft.com/dotnet/sdk:3.1
: exit status 1 
```

**Expected behavior**
The build should be success and provide dependencies.
**Screenshots**

**Any relevant environment information**
 - OS: Linux Platform builds
 
This is related to an existing issue in scanning dependencies when build cache is enabled #522

shahzzzam:
Fixed by #580

jmsalvo:
Has this fix been deployed?  I'm still seeing this issue when building using docker BUILDKIT.

shahzzzam:
Its been fixed but not deployed, Sorry for the confusion. Re-opened the issue

jmsalvo:
Thank you for the update

northtyphoon:
The fix has been deployed.

PR #580, opened 2021-05-21
[buildkit] fetch digest from remote

### LINKED ISSUE #522: Cannot track dependencies when using build cache
**Describe the bug**
When using build cache, the dependencies cannot be tracked (queried for digest).
Reason:
Because we build the image using buildx, but query the digest using docker.
`buildx` and `docker` have two separate stores. We can load the final built image into `docker` by using the `--load` flag of `buildx`, but unfortunately `buildx` does not load all the other dependent images by design.

Solution:
We need to be able to use something like `docker manifest inspect base_dependency` and grab the master digest of the image. 

Currently `docker manifest inspect` does not support getting the master manifest if the image. So can PR into docker-cli.
Or, can `acb` can use the dockerApi to make /manifest call, and parse the object to fetch sha digest.

**To Reproduce**
1. 
2. 
3. 

**Expected behavior**

**Screenshots**

**Any relevant environment information**
 - OS: [e.g. Windows]
 - Version [e.g. v1.0.0]

**Additional context**

shahzzzam:
**Purpose of the PR**

- buildkit cannot get the built image dependencies using `docker image inspect`
- So it the manifest digest from the remote registry when buildkit is used.

Fixes #522 and #574

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -162,18 +162,20 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
 	var deps []*image.Dependencies
 	for _, step := range task.Steps {
 		log.Printf("Step ID: %v marked as %v (elapsed time in seconds: %f)\n", step.ID, step.StepStatus, step.EndTime.Sub(step.StartTime).Seconds())
-		// currently we have a limitation where we cannot fetch the digest of the dependencies
-		// because we build the image using buildx, but query the digest using docker.
-		// https://github.com/Azure/acr-builder/issues/522
-		if step.UseBuildCacheForBuildStep() && runtime.GOOS == util.LinuxOS {
-			continue
-		}
+
 		if len(step.ImageDependencies) > 0 {
 			log.Printf("Populating digests for step ID: %s...\n", step.ID)
 			timeout := time.Duration(digestsTimeoutInSec) * time.Second
 			digestCtx, cancel := context.WithTimeout(ctx, timeout)
 			defer cancel()
-			if err := b.populateDigests(digestCtx, step.ImageDependencies); err != nil {
+
+			usingBuildkit := false
+			if (step.UseBuildCacheForBuildStep() && runtime.GOOS == util.LinuxOS) || step.UsesBuildkit {
+				log.Printf("Image was built using buildkit, fetching Digest from remote...")
+				usingBuildkit = true
+			}
+
+			if err := b.getPopulateDigests(digestCtx, step.ImageDependencies, usingBuildkit, task.RegistryLoginCredentials); err != nil {
 				return err
 			}
 			log.Printf("Successfully populated digests for step ID: %s\n", step.ID)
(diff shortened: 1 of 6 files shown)


shahzzzam:
ping @northtyphoon @shizhMSFT

builder/digest_remote.go
```diff
@@ -0,0 +1,72 @@
...
+
+	resolver := docker.NewResolver(opts)
+	imageRef, err := getReferencePath(ref)
+	if err != nil {
+		return err
+	}
+	_, desc, err := resolver.Resolve(ctx, imageRef)
+
+	// If the image is not pushed yet, it will not have any digest.
+	if err != nil && !strings.Contains(strings.ToLower(err.Error()), "not found") {
```
shizhMSFT, line 54: You should not test the error using strings. Instead, you should do
```go
errdefs.IsNotFound(err)
```

See https://pkg.go.dev/github.com/containerd/containerd/errdefs#IsNotFound

builder/digest_remote.go
```diff
@@ -0,0 +1,72 @@
...
+		return errors.Wrapf(err, "Failed to Resolve the reference '%s'", ref.Reference)
+	}
+	ref.Digest = desc.Digest.String()
```
shizhMSFT, line 57: What will happen if it is not found? Should we not assign `desc.Digest.String()` to `ref.Digest` if ref is not found?

In the case of `err != nil`, the value of `desc` cannot be trusted unless it is documented.

builder/digest_remote.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return nil
+}
+
+func getReferencePath(ref *image.Reference) (string, error) {
+	fullRefPath := fmt.Sprintf("%s/%s", ref.Registry, ref.Repository)
+	if ref.Tag != "" {
```
shizhMSFT, line 64: What if tag is empty? The containerd resolver will not use `latest` as the default tag. You need to specify it explicitly.

graph/step.go
```diff
@@ -347,3 +354,12 @@ func addBuildCacheOptsToCmd(domain, path, tag, originalBuildCmd string) (string,
 	}
 	return fmt.Sprintf("--load --cache-to=type=registry,ref=%s,mode=max --cache-from=type=registry,ref=%s %s", cacheImage.String(), cacheImage.String(), originalBuildCmd), nil
 }
+
+func invokesBuildkit(envs []string) bool {
+	for _, env := range envs {
+		if strings.Contains(env, BUILDKIT_ENV_VAR) {
```
shizhMSFT, line 360: What will happen for the envs beloow?
- `DOCKER_BUILDKIT=123`
- `CUSTOMIZED_DOCKER_BUILDKIT=1`
- `WHATEVER="DOCKER_BUILDKIT=1"`

builder/digest_remote.go
shahzzzam: > Should we not assign desc.Digest.String() to ref.Digest if ref is not found?

Yes, updated.

builder/digest_remote.go
shahzzzam: Agreed. Removed this function since we have already parsed the full ref while scanning dependencies.

graph/step.go
shahzzzam: Updated

builder/digest_remote.go
shahzzzam: Thanks, updated

builder/digest_remote.go
```diff
@@ -0,0 +1,92 @@
...
+			return cred.Username.ResolvedValue, cred.Password.ResolvedValue, nil
+		}
+	}
+
+	resolver := docker.NewResolver(opts)
+	imageRef, err := getReferencePath(ref)
+	if err != nil {
+		return err
+	}
+	// Note: (bug), if the final image is NOT pushed in this build,
```
shahzzzam, line 54: @shizhMSFT @northtyphoon please review this note

builder/builder.go
```diff
@@ -162,18 +162,20 @@ func (b *Builder) RunTask(ctx context.Context, task *graph.Task) error {
...
 			defer cancel()
-			if err := b.populateDigests(digestCtx, step.ImageDependencies); err != nil {
+
+			usingBuildkit := false
+			if (step.UseBuildCacheForBuildStep() && runtime.GOOS == util.LinuxOS) || step.UsesBuildkit {
+				log.Printf("Image was built using buildkit, fetching Digest from remote...")
+				usingBuildkit = true
+			}
+
+			if err := b.getPopulateDigests(digestCtx, step.ImageDependencies, usingBuildkit, task.RegistryLoginCredentials); err != nil {
```
northtyphoon, line 178: If the base image is from other foreign registry, does `RegistryLoginCredentials` include it?

builder/digest_docker.go
```diff
@@ -0,0 +1,92 @@
...
+	if reference == nil {
+		return nil
+	}
+	if reference.Digest != "" {
+		return nil
+	}
+	// refString will always have the tag specified at this point.
+	// For "scratch", we have to compare it against "scratch:latest" even though
+	// scratch:latest isn't valid in a FROM clause.
+	if reference.Reference == NoBaseImageSpecifierLatest {
```
northtyphoon, line 40: How does `digest_remote` handle the `scratch` base image?

builder/digest_remote.go
northtyphoon: I remember current implementation based on docker inspect will return empty digest if the image is not pushed. I think we need to keep the same behavior.
For the following case, I think we still can use docker inspect to query the digest of `$Registry/acb:lates`. If it's pushed, the code will return the digest. If it's not pushed, the code will return empty.
For base image, the code will switch to manifest api to get the digest for buildkit build.

builder/builder.go
northtyphoon: For example, if the customer uses managed identity to login to the foreign registry, can the code handle the case?

graph/step.go
```diff
@@ -347,3 +354,12 @@ func addBuildCacheOptsToCmd(domain, path, tag, originalBuildCmd string) (string,
 	}
 	return fmt.Sprintf("--load --cache-to=type=registry,ref=%s,mode=max --cache-from=type=registry,ref=%s %s", cacheImage.String(), cacheImage.String(), originalBuildCmd), nil
 }
+
+func invokesBuildkit(envs []string) bool {
+	for _, env := range envs {
+		if env == BUILDKIT_ENV_VAR {
```
northtyphoon, line 360: If there is space in the env string like  `DOCKER_BUILDKIT = 1`, can the code handle it?

scan/dependencies.go
```diff
@@ -147,6 +147,9 @@ func NewImageReference(imagePath string) (*image.Reference, error) {
 	if tagged, ok := ref.(reference.Tagged); ok {
 		result.Tag = tagged.Tag()
 	}
+	if digested, ok := ref.(reference.Digested); ok {
```
northtyphoon, line 150: Just curious, what is the change for? Is it to fix some related bug?

builder/digest_remote.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	// If the final image is not pushed ever, it will not have any digest.
+	if errdefs.IsNotFound(err) {
+		return nil
+	}
+
+	return errors.Wrapf(err, "Failed to Resolve the reference '%s'", ref.Reference)
+}
+
+func getReferencePath(ref *image.Reference) (string, error) {
```
northtyphoon, line 80: In the docker inspect solution, we need to special handle library images, eg, nginx, ubuntu, because of the `library` prefix. Does the remote inspect work for these library image?

graph/step.go
shahzzzam: if there's space in the env string, then the Task itself will fail before running docker build, since I believe we just paste all the Env passed to `docker build` and any space in `--env` leads to error.

scan/dependencies.go
shahzzzam: This change is for dockerfiles like:

```Dockerfile
FROM busybox@sha256:f3cfc9d0dbf931d3db4685ec659b7ac68e2a578219da4aae65427886e649b06b
RUN echo "hello" > world.txt
```

As you see, `busybox` already is digested, there is no need to `docker inspect` it or call Manifest API to resolve the digest, we already resolve it while scanning dependencies.

That's why I added the following to return fast in `populateDigest()` method:

```go
if ref.Digest != ""  {
    return nil	
}
```

builder/builder.go
shahzzzam: Yes, they are resolved credentials (opaque, KV, and MSI)

builder/digest_remote.go
shahzzzam: Great idea, thanks.

builder/digest_docker.go
shahzzzam: Added that

graph/step.go
shahzzzam: For ex.

```sh
2021/06/01 20:35:23 Step args: /bin/sh, -c, docker run --rm --network acb_default_network --name build --volume source:/workspace --volume /var/run/docker.sock:/var/run/docker.sock \
       --volume home:/acb/home --env HOME=/acb/home --env DOCKER_BUILDKIT = 1 --workdir /workspace docker build -t samashahtest.azurecr.io/acb:blah -f test_Dockerfile .
2021/06/01 20:35:23 Launching container with name: build
docker: invalid reference format.
See 'docker run --help'.

```

builder/digest_remote.go
shahzzzam: That only needs to be done because the `docker inspect` returns an output like:
```
["busybox@sha256:b5fc1d7b2e4ea86a06b0cf88de915a2c43a99a00b6b3c0af731e5f4c07ae8eff","busybox@sha256:f3cfc9d0dbf931d3db4685ec659b7ac68e2a578219da4aae65427886e649b06b"]
```

We just remove the prefix library/ if we attached it.

We don't need to do any of that in remote

northtyphoon (approved):

jmsalvo:
Has this fix been deployed? I'm still seeing this issue when building using docker BUILDKIT.

PR #590, opened 2022-01-12
install git lfs on system level

northtyphoon:
The change avoids losing git lfs setting if the default home folder is overwritten.

fix#589


rosanch (approved):
LGTM

PR #592, opened 2022-04-12
update dependency

northtyphoon:
**Purpose of the PR**

1. Update azure sdk
2. Update golang tool to 1.17


juliusl (approved):

estebanreyl:
PR Looks good to me. The only question I do have is why we don't currently commit the vendor folder. Since Go dependencies are independently managed (as in there is no global artifact store for them an owner for them can remove a commit, update naming or even delete the dependency which would render the vendor building unfeasible making acr-builder not work. In addition to this go 1.17 adds vendor/modules.txt which adds further vendoring information.

northtyphoon:
> PR Looks good to me. The only question I do have is why we don't currently commit the vendor folder. Since Go dependencies are independently managed (as in there is no global artifact store for them) an owner for them can remove a commit, update naming or even delete the dependency which would render the vendor building unfeasible making acr-builder not work. In addition to this go 1.17 adds vendor/modules.txt which adds further vendoring information.

Thanks. We will update vendor again in the coming oras-go refactoring

PR #598, opened 2022-04-14
migrate to oras-go/v2

northtyphoon:
**Purpose of the PR**

- migrate artifact download to oras-go/v2


scan/context.go
```diff
@@ -121,26 +123,39 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
...
-	})
-	resolver := docker.NewResolver(docker.ResolverOptions{
-		PlainHTTP: false,
-		Hosts: docker.ConfigureDefaultRegistries(
-			docker.WithAuthorizer(authorizer),
-		),
-	})
-	fileStore := content.NewFileStore(s.destinationFolder)
-	defer fileStore.Close()
+	src, err := remote.NewRepository(registryArtifact)
```
Wwwsylvia, line 126: What would the `registryArtifact` string look like?

scan/context.go
```diff
@@ -121,26 +123,39 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
 }
 
 func (s *Scanner) getContextFromRegistry(ctx context.Context, registryArtifact string) (err error) {
```
Wwwsylvia, line 125: How do we easily verify it?
I'm a bit confused by the command description... Do we have an example for `scan`?
image

scan/context.go
northtyphoon: myregistry.azurecr.io/myartifact:mytag

Wwwsylvia: Thanks!

scan/context.go
northtyphoon: I guess the scan module might give you an impression that the corresponding command is `scan`. It's also used in `download` command. To test the code, you can sync the code and run 
```
# build the binary into bin folder
make build
# download the oci artifact to default temp folder
cd bin
./acb download oci://binducentralus.azurecr.io/acb-oras-sample:arm
# verify the downloaded file
tree temp/arm
```

Wwwsylvia: Ok cool, it works! Thank you!

northtyphoon: Forgot to mention, here is the command to test authentication with `--credential` parameter

```
./acb download --credential '{"registry":"binducentralus.azurecr.io","username":"binducentralus","userNameProviderType":"opaque","password":"replace-the-password","passwordProviderType":"opaque"}' oci://binducentralus.azurecr.io/acb-oras-sample:arm
```

scan/context.go
```diff
@@ -121,26 +123,39 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
...
+		},
+		Cache: auth.DefaultCache,
+		Credential: func(ctx context.Context, registry string) (auth.Credential, error) {
+			// If no matching credential found, attempt an anonymous pull
+			if s.credentials[registry] == nil {
+				return auth.EmptyCredential, nil
+			}
+
+			return auth.Credential{
+				Username: s.credentials[registry].Username.ResolvedValue,
```
akashsinghal, line 143: Looks like I forgot to add a nil check for the `Username` and `Password` field when I wrote the code a originally. Do you think we need to add it? I guess maybe it hasn’t been a problem so far?

northtyphoon: It doesn't look it can be nil

https://github.com/Azure/acr-builder/blob/d448eceff8c1117524f6d8bcfdec01ad54d05e68/graph/task.go#L441

Wwwsylvia (commented):
LGTM

scan/context.go
akashsinghal: Ahh ok nevermind then.

akashsinghal (approved):
LGTM

PR #614, opened 2022-07-09
check exchange api response and return the appropriate error

### LINKED ISSUE #613: The error message is misleading if identity doesn't have permission on the registry
If the identity doesn't have the permission to access the registry, the code will display the following error message

`failed to login, ran out of retries: failed to set docker credentials: Error: Cannot perform an interactive login from a non TTY device`

It's because current code didn't check the response status code of ACR exchange api. The code ends up passing an empty password string to docker login command.

https://github.com/Azure/acr-builder/blob/c94aeab189108160b2c60554cd82b63a1010e4bc/tokenutil/token_util.go#L54

northtyphoon:
**Purpose of the PR**

- properly handle status code from ACR exchange api

Fixes #613

### DIFF
--- a/tokenutil/token_util.go
+++ b/tokenutil/token_util.go
@@ -47,25 +47,29 @@ func GetRegistryRefreshToken(registry, resourceID, clientID string) (string, err
 
 	req, err := http.NewRequest("POST", exchangeURL, strings.NewReader(v.Encode()))
 	if err != nil {
-		return "", errors.Wrap(err, "unable to create an HTTP post request to exchangeAPI")
+		return "", errors.Wrap(err, "unable to create the request to get ACR refresh token")
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
 	response, err := client.Do(req)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


tokenutil/token_util.go
```diff
@@ -47,25 +47,29 @@ func GetRegistryRefreshToken(registry, resourceID, clientID string) (string, err
...
 	if err != nil {
-		return "", errors.Wrap(err, "unable to create an HTTP post request to exchangeAPI")
+		return "", errors.Wrap(err, "unable to create the request to get ACR refresh token")
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
 	response, err := client.Do(req)
 	if err != nil {
-		return "", errors.Wrap(err, "unable to make post request to exchangeAPI")
+		return "", errors.Wrap(err, "unable to make the request to get ACR refresh token")
```
estebanreyl, line 56: unable to make sounds a bit odd. Should we do "unable to send..." instead, it seems more clear

estebanreyl (approved):
LGTM

PR #621, opened 2022-09-13
build: switch to build with local vendor sources

northtyphoon:
**Purpose of the PR**

- the change switch go build to use local vendor folder for all dependencies


estebanreyl (approved):
LGTM, thanks Bin :)

PR #646, opened 2023-07-12
fix: switch docker build to use curl command to download git-lfs

northtyphoon:
**Purpose of the PR**

- wget is not available on mariner-based image, switch to curl command


estebanreyl (approved):
LGTM

Issue #653, opened 2023-10-20
 Intermittent failure to download build context tarball from azure storage blob: unexpected EOF


northtyphoon:
**Describe the bug**
acr download <uri of tarball storage blob> sometimes failed saying "unexpected EOF"

Suggestion:
Add retry to avoid intermittent network issue

yuehaoliang-microsoft:
Suspect that the "unexpected EOF" error may be originating from the `archive.Untar()` function, but not entirely certain. So added some logs within "acb download" to pinpoint the source of the error. The customer has reported that this error consistently occurs for a particular task, so retrying may not work.

This might also be related to something when the Azure CLI "az acr build" packaging and uploading the tarball.

yuehaoliang-microsoft:
I've identified that the issue is primarily related to the file size, and the threshold for encountering the issue is approximately around 500 MB. See examples:
 
```
Packing source code into tar to upload...
Excluding '.git' based on default ignore rules
Uploading archived source code from 'C:\Users\haoliang\AppData\Local\Temp\build_archive_726c8deede7c45c28a20856a411acfe2.tar.gz'...
Sending context (**435.320 Mi**B) to registry: yuehaoliangwuscr...
Queued a build with ID: ydh
Waiting for an agent...
2023/11/02 09:50:23 Downloading source code...
**2023/11/02 09:50:30 Finished downloading source code**
```
 
```
Packing source code into tar to upload...
Excluding '.git' based on default ignore rules
Uploading archived source code from 'C:\Users\haoliang\AppData\Local\Temp\build_archive_a28c371755874a188a6566aae9355942.tar.gz'...
Sending context (**686.414 MiB**) to registry: yuehaoliangwuscr...
Queued a build with ID: ydg
Waiting for an agent...
2023/11/02 09:47:07 Downloading source code...
**Run ID: ydg failed after 3s. Error: failed to download context. Please check if the URL is incorrect. If it has credentials, please check if they are expired**
Run failed
```
 
```
Packing source code into tar to upload...
Excluding '.git' based on default ignore rules
Uploading archived source code from 'C:\Users\haoliang\AppData\Local\Temp\build_archive_41756eb7d9664758944035e2012842cd.tar.gz'...
Sending context (**919.603 MiB**) to registry: yuehaoliangwuscr...
Queued a build with ID: ydf
Waiting for an agent...
2023/11/02 09:37:50 Downloading source code...
[...]

PR #654, opened 2023-10-25
add troubleshooting logs in acb download

yuehaoliang-microsoft:
**Purpose of the PR**
Observed some "unexpected EOF" when downloading the context, so add more logs for troubleshooting.

time="2023-10-05T16:49:50Z" level=debug msg="Output from downloading: unexpected EOF\n"
time="2023-10-05T16:49:50Z" level=warning msg="Failed to download source context. Error: exit status 1"

I think "unexpected EOF" is likely originated from _**archive.Untar()**_. This method will throw "unexpected EOF" if the content is invalid or corrupted. If so, these logs may not be very helpful. Any suggestions?

Fixes #

### DIFF
--- a/cmd/acb/commands/download/download.go
+++ b/cmd/acb/commands/download/download.go
@@ -63,6 +63,7 @@ var Command = cli.Command{
 		// Add all creds provided by the user in the --credential flag
 		credentials, err := graph.CreateRegistryCredentialFromList(creds)
 		if err != nil {
+			log.Println("Failed to add credentials")
 			return err
 		}
 
@@ -75,16 +76,20 @@ var Command = cli.Command{
 		if util.IsRegistryArtifact(downloadCtx) {
 			registryLoginCredentials, err = graph.ResolveCustomRegistryCredentials(ctx, credentials)
 			if err != nil {
+				log.Println("Failed to resolve credentials")
 				return err
 			}
 		}
 
 		scanner, err := scan.NewScanner(pm, downloadCtx, "", destination, nil, nil, "", registryLoginCredentials)
 		if err != nil {
+			log.Println("Failed to create new scanner")
 			return err
 		}
+
 		workingDir, sha, branch, err := scanner.ObtainSourceCode(ctx, downloadCtx)
 		if err != nil {
+			log.Println("Failed to obtain source code")
 			return err
 		}
 
--- a/cmd/acb/commands/download/download.go
+++ b/cmd/acb/commands/download/download.go
@@ -60,6 +60,8 @@ var Command = cli.Command{
 			return errors.New("download requires context to be provided, see download --help")
 		}
 
+		log.Println("Downloading context")
+
 		// Add all creds provided by the user in the --credential flag
 		credentials, err := graph.CreateRegistryCredentialFromList(creds)
 		if err != nil {
--- a/scan/context.go
+++ b/scan/context.go
@@ -63,7 +63,7 @@ func (s *Scanner) getContext(ctx context.Context, scContext string) (workingDir 
 		workingDir, err := s.getContextFromGitURL(scContext)
 		return workingDir, err
 	} else if isURL {
-		fmt.Printf("Getting context from URL %s\n", scContext)
+		fmt.Println("Getting context from URL")
 		err := s.getContextFromURL(scContext)
 		return s.destinationFolder, err
 	} else if isRegistryArtifact {
@@ -92,7 +92,7 @@ func (s *Scanner) getContextFromURL(remoteURL string) (err error) {
 		var response *http.Response
 		response, err = getWithStatusError(remoteURL)
 		if err != nil {
-			return errors.Wrapf(err, "unable to download remote context from %s", remoteURL)
+			return errors.Wrap(err, "unable to download remote context")
 		}
 
 		fmt.Printf("Read context with status code %d\n", response.StatusCode)
--- a/scan/context.go
+++ b/scan/context.go
@@ -60,16 +60,20 @@ func (s *Scanner) getContext(ctx context.Context, scContext string) (workingDir 
 	}
 
 	if isSourceControlURL {
+		fmt.Println("Getting context from Git URL")
 		workingDir, err := s.getContextFromGitURL(scContext)
 		return workingDir, err
 	} else if isURL {
+		fmt.Println("Getting context from URL")
 		err := s.getContextFromURL(scContext)
 		return s.destinationFolder, err
 	} else if isRegistryArtifact {
+		fmt.Println("Getting context from registry")
 		err := s.getContextFromRegistry(ctx, util.TrimArtifactPrefix(scContext))
 		return s.destinationFolder, err
 	}
 
+	fmt.Println("Context is local")
 	return scContext, nil
 }
 
--- a/scan/context.go
+++ b/scan/context.go
@@ -93,6 +93,8 @@ func (s *Scanner) getContextFromURL(remoteURL string) (err error) {
 		return errors.Wrapf(err, "unable to download remote context from %s", remoteURL)
 	}
 
+	fmt.Printf("Read context of %d bytes", response.ContentLength)
+
 	// TODO: revamp streaming, for now just pipe to buf and discard it.
 	var buf bytes.Buffer
 	progressOutput := streamformatter.NewProgressOutput(&buf)
@@ -116,7 +118,11 @@ func (s *Scanner) getContextFromReader(r io.Reader) (err error) {
 	}
 
 	if dockerbuild.IsArchive(magic) {
+		fmt.Println("starting to untar context")
 		err = archive.Untar(buf, s.destinationFolder, nil)
+		if err != nil {
+			return errors.Wrap(err, "failed to untar context")
+		}
 	}
 
 	return err
--- a/scan/context.go
+++ b/scan/context.go
@@ -4,6 +4,7 @@
 package scan
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
@@ -19,6 +20,7 @@ import (
 	"oras.land/oras-go/v2/registry/remote/auth"
 
 	"github.com/Azure/acr-builder/util"
+	dockerbuild "github.com/docker/cli/cli/command/image/build"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/docker/docker/pkg/progress"
 	"github.com/docker/docker/pkg/streamformatter"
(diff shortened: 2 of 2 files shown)


cmd/acb/commands/download/download.go
```diff
@@ -63,6 +63,7 @@ var Command = cli.Command{
 		// Add all creds provided by the user in the --credential flag
 		credentials, err := graph.CreateRegistryCredentialFromList(creds)
 		if err != nil {
+			log.Println("Failed to add credentials")
```
northtyphoon, line 66: does the caller log the err?

cmd/acb/commands/download/download.go
```diff
@@ -75,16 +76,20 @@ var Command = cli.Command{
 		if util.IsRegistryArtifact(downloadCtx) {
 			registryLoginCredentials, err = graph.ResolveCustomRegistryCredentials(ctx, credentials)
 			if err != nil {
+				log.Println("Failed to resolve credentials")
```
northtyphoon, line 79: does the caller log the err?

cmd/acb/commands/download/download.go
```diff
@@ -75,16 +76,20 @@ var Command = cli.Command{
...
 				return err
 			}
 		}
 
 		scanner, err := scan.NewScanner(pm, downloadCtx, "", destination, nil, nil, "", registryLoginCredentials)
 		if err != nil {
+			log.Println("Failed to create new scanner")
```
northtyphoon, line 86: does the caller log the err?

cmd/acb/commands/download/download.go
```diff
@@ -75,16 +76,20 @@ var Command = cli.Command{
...
 			return err
 		}
+
 		workingDir, sha, branch, err := scanner.ObtainSourceCode(ctx, downloadCtx)
 		if err != nil {
+			log.Println("Failed to obtain source code")
```
northtyphoon, line 92: does the caller log the error?

northtyphoon:
https://github.com/Azure/acr-builder/issues/653

cmd/acb/commands/download/download.go
yuehaoliang-microsoft: I think so. worker.go has

```
log.Debugf("Output from downloading: %s", downloadOutput)
if err != nil {
    log.Warningf("Failed to download source context. Error: %s", err)
}
```
where `downloadOutput` contains stdOut and stdErr of the command run. The added log should be included in it.

Eg.

time="2023-10-05T16:49:50Z" level=debug msg="Output from downloading: unexpected EOF\n"
time="2023-10-05T16:49:50Z" level=warning msg="Failed to download source context. Error: exit status 1"

time="2023-10-23T06:16:39Z" level=debug msg="Output from downloading: fatal: not a git repository (or any parent up to mount point /)\nStopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).\nfatal: not a git repository (or any parent up to mount point /)\nStopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).\n2023/10/23 06:16:39 Download complete, working directory: /workspace\n2023/10/23 06:16:39 CommitID and Branch information: {\"commitID\":\"\",\"branch\":\"\"}\n"

scan/context.go
```diff
@@ -93,6 +93,8 @@ func (s *Scanner) getContextFromURL(remoteURL string) (err error) {
 		return errors.Wrapf(err, "unable to download remote context from %s", remoteURL)
 	}
 
+	fmt.Printf("Read context of %d bytes", response.ContentLength)
```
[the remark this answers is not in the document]
yuehaoliang-microsoft: Yeah, I think check the file size may be helpful.

I've tried that the response should provide the information:
```
url := "https://golang.google.cn/dl/go1.21.3.linux-amd64.tar.gz"
response, _ := http.Get(url)
log.Printf("StatusCode is %d", response.StatusCode)
log.Printf("ContentLength is %d", response.ContentLength)
```

2023/10/27 11:29:14 StatusCode is 200
2023/10/27 11:29:14 ContentLength is 66641773

ls -l
66641773 Oct 11 00:32 go1.21.3.linux-amd64.tar.gz

yuehaoliang-microsoft:
> Can we add a retry in case of Intermittent download error?
> 
> #653

Sure, I will add a retry.

I also wanted to wish some luck in seeing how [`docker import`](https://docs.docker.com/engine/reference/commandline/import/) is written for streaming, but it turns out we had already adapted the code from Moby, [moby/api/server/router/image/image_routes.go](https://github.com/moby/moby/blob/master/api/server/router/image/image_routes.go#L130-L141). I still see some differences in the way it's written, and I plan to try it out.

scan/context.go
```diff
@@ -185,12 +176,12 @@ func getWithStatusError(url string) (resp *http.Response, err error) {
 		// If the status code is 4xx then read the body and return an error
 		// If the status code is a 5xx then check the body on the final attempt and return an error.
 		if resp.StatusCode < 500 || attempt == maxRetries-1 {
-			body, bodyReadErr := ioutil.ReadAll(resp.Body)
```
yuehaoliang-microsoft, line 188: `ioutil` is deprecated as of Go 1.16, `io` should be preferred for the same functionality.

scan/context.go
```diff
@@ -170,12 +174,14 @@ func (s *Scanner) getContextFromRegistry(ctx context.Context, registryArtifact s
 // - The response is 5xx
 func getWithStatusError(url string) (resp *http.Response, err error) {
 	attempt := 0
-
 	for attempt < maxRetries {
 		resp, err = http.Get(url)
 		if err != nil {
+			time.Sleep(util.GetExponentialBackoff(attempt))
+			attempt++
 			continue
```
yuehaoliang-microsoft, line 182: It looks missing `attemp++` here? if `http.Get(url)` has `err`, then it will run an infinity loop.

scan/context.go
```diff
@@ -88,35 +87,40 @@ func (s *Scanner) getContextFromGitURL(gitURL string) (contextDir string, err er
 }
 
 func (s *Scanner) getContextFromURL(remoteURL string) (err error) {
-	response, err := getWithStatusError(remoteURL)
-	if err != nil {
-		return errors.Wrapf(err, "unable to download remote context from %s", remoteURL)
-	}
+	attempt := 0
```
yuehaoliang-microsoft, line 90: `http.Response.Body` is a one-time stream. If we want to retry the untar step, we need to do `http.Get()` again.

scan/context.go
```diff
@@ -63,12 +59,15 @@ func (s *Scanner) getContext(ctx context.Context, scContext string) (workingDir
 	}
 
 	if isSourceControlURL {
+		fmt.Println("Getting context from Git URL")
 		workingDir, err := s.getContextFromGitURL(scContext)
 		return workingDir, err
 	} else if isURL {
+		fmt.Printf("Getting context from URL %s\n", scContext)
```
yuehaoliang-microsoft, line 66: I see the URL is actually logged arround line 95 when `http.Get()` has an error. So, just want to confirm it should be ok to log this info?

scan/context.go
```diff
@@ -88,35 +87,40 @@ func (s *Scanner) getContextFromGitURL(gitURL string) (contextDir string, err er
...
+		err = s.getContextFromReader(rc)
+		if err == nil {
+			return nil
+		}
+
+		attempt++
+		time.Sleep(util.GetExponentialBackoff(attempt))
 	}
 
-	if dockerbuild.IsArchive(magic) {
```
yuehaoliang-microsoft, line 118: Would like to discuss is it necessary to peek the magic bytes and check whether the content is archive? If its content is not valid, we can let `archive.Untar()` return the error.

scan/context.go
```diff
@@ -88,35 +87,40 @@ func (s *Scanner) getContextFromGitURL(gitURL string) (contextDir string, err er
 }
 
 func (s *Scanner) getContextFromURL(remoteURL string) (err error) {
-	response, err := getWithStatusError(remoteURL)
-	if err != nil {
-		return errors.Wrapf(err, "unable to download remote context from %s", remoteURL)
```
yuehaoliang-microsoft, line 93: Could the URL be logged? Seems it contains SAS.

yuehaoliang-microsoft:
After the change, an example log will be:

time="2023-11-01T07:30:54Z" level=debug msg="Output from downloading: 2023/11/01 07:30:53 Downloading context\nGetting context from URL\nRead context with status code 200\nRead context of 4762 bytes\nstarting to untar context\nfatal: not a git repository (or any parent up to mount point /)\nStopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).\nfatal: not a git repository (or any parent up to mount point /)\nStopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).\n2023/11/01 07:30:53 Download complete, working directory: /workspace\n2023/11/01 07:30:53 CommitID and Branch information: {\"commitID\":\"\",\"branch\":\"\"}\n"

scan/context.go
```diff
@@ -88,35 +87,40 @@ func (s *Scanner) getContextFromGitURL(gitURL string) (contextDir string, err er
...
+	attempt := 0
+	for attempt < maxRetries {
+		var response *http.Response
+		response, err = getWithStatusError(remoteURL)
+		if err != nil {
+			return errors.Wrap(err, "unable to download remote context")
+		}
 
-	// TODO: revamp streaming, for now just pipe to buf and discard it.
-	var buf bytes.Buffer
```
yuehaoliang-microsoft, line 97: Can just use `io.Discard`.

scan/context.go
northtyphoon: The original code will skip untar and return. The pr changes the logic. Do you think it's related to the netowrk issue? If not, I suggest we keep the original logic in this round.

scan/context.go
northtyphoon: good catch

scan/context.go
northtyphoon: Is the behavior (what we eventually log) the same as before after the pr changes to `io.Discard`?

scan/context.go
northtyphoon: It's very risky to log full URL as it can contains sensitive information.

scan/context.go
yuehaoliang-microsoft: The progress bar is not printed before this PR, because the `buf` is not used.
```
var buf bytes.Buffer
progressOutput := streamformatter.NewProgressOutput(&buf)
```
The behavior remains the same if change to 
```
progressOutput := streamformatter.NewProgressOutput(io.Discard)
```

scan/context.go
yuehaoliang-microsoft: Sure, rolled back the logic to checking archive before doing untar. I don't think it's related to the issue, just trying to make the streaming a little bit simpler.

northtyphoon (approved):
LGTM

PR #661, opened 2023-11-08
fix a bug in download stream

yuehaoliang-microsoft:
**Fix a bug left in my last PR**

I forgot to revert `r` to `buf`, resulting in "failed to untar context: archive/tar: invalid tar header".
image


northtyphoon (approved):
LGTM

PR #677, opened 2024-04-22
feat: sensor potential sensitive data in messages

yuehaoliang:
**Purpose of the PR**
- We discovered a scenario when `git fetch` had such stdout/stderr:
```
git: 'credential-cache' is not a git command. See 'git --help'.\nremote: Invalid username or password.
fatal: Authentication failed for 'https://dummy:<PAT>@github.com/<USER>/<REPO>.git/'
: exit status 128\
```

Fixes
- Check whether the error message contains the PAT, if so, censor it from the message before logging it.


scan/git.go
```diff
@@ -306,3 +307,10 @@ func git(args ...string) ([]byte, error) {
 func isGitTransport(str string) bool {
 	return urlutil.IsURL(str) || strings.HasPrefix(str, "git://") || strings.HasPrefix(str, "git@")
 }
+
```
northtyphoon, line 310: I think we also need to handle `visualstudio.com` and `dev.azure.com` for Azure DevOps

yuehaoliang: Removed 'github' from the pattern.

wju-MSFT (approved):
LGTM

PR #684, opened 2024-07-03
fix: allow escaped version key

### LINKED ISSUE #685: acr-builder can not find tasks.yaml version if the version key has been escaped
**Describe the bug**
**See Title**

**To Reproduce**
Use `"version": v1.1.0` instead of `version: v1.1.0`

**Expected behavior**

Works

**Additional context**

Tools like terraform will escape yaml keys when encoding yaml content which can lead to unexpected behavior when users try to deploy their tasks.yaml through terraform.

juliusl:
…because terraform will always encode escape yaml property keys

**Purpose of the PR**

- Today we do not allow `"version": v1.1.0` when finding the tasks.yaml version during preprocess. This change allows the preprocessor to find the version key if the version key has been escaped.

Fixes #685

### DIFF
--- a/graph/preprocessor_test.go
+++ b/graph/preprocessor_test.go
@@ -626,12 +626,112 @@ version:   'v1.1.0'`,
 version:   "v1.1.0'`,
 			"v1.1.0",
 		},
+		{
+			`       
+"version"`,
+			"",
+		},
+		{
+			` # task.yml file
+# should be skipped
+
+"version"  : v1.1.0    
+`,
+			"v1.1.0",
+		},
+		{
+			`
+build: something
+"version": v1.1.0`,
+			"",
+		},
+		{
+			`       
+"version"`,
+			"",
+		},
+		{
+			`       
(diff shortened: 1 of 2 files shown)


northtyphoon (approved):
LGTM

graph/preprocessor_test.go
```diff
@@ -626,12 +626,112 @@ version:   'v1.1.0'`,
...
+			"v1.1.0",
+		},
+		{
+			`       
+"version":   'v1.1.0'`,
+			"v1.1.0",
+		},
+		{
+			`       
+"version":   "v1.1.0'`,
```
estebanreyl, line 676: So from following this it looks like we end up supporting a fair bit of invalid inputs for parsing but since the goal of this function is to get the version, not validate that the yaml itself is valid I think this is fine

estebanreyl:
LGTM

estebanreyl (approved):

PR #687, opened 2024-07-23
fix: find version checks entire file and not just first line

juliusl:
**Purpose of the PR**

- Today find version only checks the first line for the version. This allows the whole file to be checked instead.

Fixes #


estebanreyl (approved):
LGTM

PR #710, opened 2025-07-04
fix: pre-run windows server 2019 containers to prevent startup issues

stevenhvtran:
**Purpose of the PR**
- [April 16, 2025—KB5059091 (OS Build 17763.7249) Out-of-band - Microsoft Support](https://support.microsoft.com/en-us/topic/april-16-2025-kb5059091-os-build-17763-7249-out-of-band-328ff8b5-7c6b-4f06-95cc-67fbd18b3ffb) introduced a bug where-in the first time a Windows Server 2019 Hyper-V container is run in Windows 2022, it has a chance of failing. This fix will silently run the container once before the user runs it, to prevent the issue

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -344,16 +344,17 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step, credentials []*
 			}
 
 			if b.debug {
-				log.Printf("Pre-run args: %v\n", strings.Join(args, ", "))
+				log.Printf("Pre-run args: %v\n", strings.Join(preRunArgs, ", "))
 			}
 
-			// Silently run the command to not confuse the user
-			err := b.procManager.Run(ctx, preRunArgs, nil, nil, nil, "")
+			// Silently run the command to not confuse the user. Only expose error in debug mode.
+			var stdErrBuf bytes.Buffer
+			err := b.procManager.Run(ctx, preRunArgs, nil, nil, &stdErrBuf, "")
 			if b.debug {
 				if err != nil {
-					log.Println("Prevented a crash by running a pre-run container")
+					log.Printf("Pre-run ran with error: %s\n", stdErrBuf.String())
 				} else {
-					log.Printf("Pre-run ran without issues")
+					log.Printf("Pre-run ran without issues\n")
 				}
 			}
 		}
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -328,6 +328,31 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step, credentials []*
 	stepCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
+	// TODO(transteven): Remove this once Windows Server 2019 Hyper-V can run on Windows Server 2022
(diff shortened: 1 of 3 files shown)


stevenhvtran:
@microsoft-github-policy-service agree company="Microsoft"

stevenhvtran:
Example run with debug logs enabled:
image

builder/builder.go
```diff
@@ -328,6 +331,15 @@ func (b *Builder) runStep(ctx context.Context, step *graph.Step, credentials []*
 	stepCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
+	// TODO(transteven): Remove this once Windows Server 2019 Hyper-V can run on Windows Server 2022
+	// without any startup issues.
+	// The current workaround is to run a throwaway container to ensure subsequent runs succeed.
+	if runtime.GOOS == util.WindowsOS && (step.Isolation == "" || step.Isolation == "hyperv") {
+		if parseImageNameFromArgs(step.Cmd) == "mcr.microsoft.com/windows/servercore:ltsc2019" || step.ContainsImageDependency("mcr.microsoft.com/windows/servercore:ltsc2019") {
+			once.Do(func() { b.preRunWindowsContainer(stepCtx, step) })
```
northtyphoon, line 339: Why do we need the `once.Do` logic?

builder/builder.go
```diff
@@ -404,6 +416,33 @@ func parseImageNameFromArgs(cmdArgs string) string {
 	return cmdArgs[:idx]
 }
 
+// preRunWindowsContainer runs a Windows Server 2019 Hyper-V container to ensure subsequent runs succeed.
+func (b *Builder) preRunWindowsContainer(ctx context.Context, step *graph.Step) {
+	preRunArgs := []string{
+		"docker",
```
northtyphoon, line 422: in build step, we add `--pull` option to make sure base image is update to date, if the workaround is for the specific image, then the warmup needs to use `--pull always` to download the latest.

builder/builder.go
stevenhvtran: This will prevent us from running the warmup windows container multiple times if the user has a multi-stage dockerfile or has multiple task commands that use the windows/servercore image

stevenhvtran: Tested an example with that uses the windows image switch and validated that it indeed only runs once:
image

builder/builder.go
stevenhvtran: added now

builder/builder.go
```diff
@@ -404,6 +416,33 @@ func parseImageNameFromArgs(cmdArgs string) string {
...
+		"run",
+		"--rm",
+		"--name", step.ID + "_prerun",
+		"--isolation", "hyperv",
+		"mcr.microsoft.com/windows/servercore:ltsc2019",
```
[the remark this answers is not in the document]
stevenhvtran: I've parameterized the image name now. I've also not been able to reproduce the issue with any other Windows 2019 images - just Windows Server Core.

PR #722, opened 2026-02-14
fix: update golang builder image

northtyphoon:
**Purpose of the PR**

- update golang builder image to 1.25.7

### DIFF
--- a/azure-pipelines-windows.yml
+++ b/azure-pipelines-windows.yml
@@ -32,3 +32,13 @@ steps:
     go test ./... 
   workingDirectory: '$(ModulePath)'
   displayName: 'Test'
+
+- script: |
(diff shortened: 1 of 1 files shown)


azure-pipelines-windows.yml
```diff
@@ -28,8 +28,17 @@ steps:
...
 
-
 - script: |
     go test ./... 
   workingDirectory: '$(ModulePath)'
-  displayName: 'Test'
\ No newline at end of file
+  displayName: 'Test'
+
+- script: |
```
estebanreyl: I see, so this is to validate that the docker builds will work?

estebanreyl (approved):
LGTM

PR #743, opened 2026-09-06
chore: remove github.com/docker/docker dependency

northtyphoon:
Replace the small pieces borrowed from github.com/docker/docker with local equivalents so the module no longer needs to vendor the full docker/docker package:

- util: reimplement IsURL/IsGitURL (previously docker/docker/builder/remotecontext/urlutil) as small local helpers.
- scan/git.go: use the local util.IsURL instead of urlutil.IsURL.
- scan/context.go: drop docker/docker/pkg/progress and pkg/streamformatter. The progress reader wrapped the response body but reported progress to io.Discard, so it was a no-op; read response.Body directly instead.
- go.mod/go.sum/vendor: run go mod tidy && go mod vendor to drop the now-unused docker/docker dependency (and its now-unreferenced transitive deps).

Also add unit tests covering the git URL protocols (git://, git@, github.com/ prefix) and the IsURL helper, plus clarifying comments.

**Purpose of the PR**

- CVE-2026-41567, CVE-2026-42306, CVE-2026-33997, CVE-2026-41568


huanwu (approved):

estebanreyl (approved):
LGTM

